PR #57, opened 2020-12-02
fix: do not use vertical var for both directions in TOC styling

Simek:
Currently TOC styles uses `--ifm-toc-padding-vertical` variable to set both, vertical and horizontal values of `li` element margin. 

This PR fixes this issue and replaces the `-ifm-toc-padding-left` using newly added `-ifm-toc-padding-horizontal`.

The changes should not affect the styling in any way.


yangshun (commented):
What's the motivation for this change? I'd like to minimize variable renaming if possible (but better earlier than later)

yangshun (approved):

PR #56, opened 2020-12-02
chore: cleanup package.json files, add IDEA related ignore

Simek:
This small PR removes the unnecessary and potentially problematic fields from `package.json` files, fixes the [repository format](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#repository) in the main file and adds a IntelliJ IDEA project/cache directory to the `.gitignore`.

PR #58, opened 2020-12-02
fix: use correct extension for PCSS files, update Demo

Simek:
Currently the Infima generate CSS files using very unusually configured `gulp` pipeline which allows SCSS syntax in CSS files and transforms it without utilizing the `gulp-sass` or `node-sass` directly. Also the `import` statement is handled in a weird, non-SASS/SCSS way (in SASS/SCSS all partials should begin with `_` and only files which are generating CSS should have normal names, in Infima example: only `default.scss` and `default-dark.scss` files should not be prefixed with underscore).

This PR is my attempt to improve the situation a bit, unfortunately due to the problems I have encountered with `postcssEasyImport` forced me to temporary copy the files as CSS and removed them after the build. To cleanup up thing a bit I have also move the `theme` folder from the styles yo the Infima main directory (so SCSS partials can be separated from files used to generate CSS).

~~I have tried to replace the outdated `postcss-easy-import` with `postcss-import` but the switch yields another build problems related to the SCSS imports.~~

~~Changes in this PR are only a mitigations, not proper fixes.~~ My long term recommendation is to migrate completely from `gulp` to other future-proof tool like `webpack` or `parcel` ~~and use SASS preprocessor, and then PostCSS.~~

After the changes both commands from `core` package, `gulp build` and `gulp` are working correctly for me.

And yes, I know that I can overwrite the syntax settings for the project in my IDE, but it's not a real fix. 😉 

### Preview of issue
Screenshot 2020-12-02 at 17 27 14


lex111 (commented):
But we are using PostCSS, so using the Sass extension is semantic wrong. Back then, `pcss` was the correct extension for style files.

Simek:
> But we are using PostCSS, so using the Sass extension is semantic wrong.

I can agree that this might not be advised this days, as it used to be before, but it was just a small point in the comment. I was able to make a bit more research into the topic and I have found few nice articles, and in particular this one:
* https://matti.dev/post/from-sass-to-postcss-ten-minutes

which ensured me that SCSS files can be used currently without a stand alone preprocessor and help me understand the import issue that I was having after initial migration to `postcss-import`.

 I was able to convert all files to `scss`, fix imports, and roll back most of my additions. 

> Back then, `pcss` was the correct extension for style files.

Current setup should handle well changing extension from `.scss` to `.pcss`, so feel free to do it, if you think it suits the files better (I'm not that experienced with PostCSS after all).

packages/postcss-preset-infima/package.json
```diff
@@ -22,7 +22,7 @@
     "postcss-combine-duplicated-selectors": "^9.1.0",
     "postcss-css-variables": "^0.12.0",
     "postcss-each": "^0.10.0",
-    "postcss-easy-import": "^3.0.0",
+    "postcss-import": "^12.0.1",
```
Simek, line 25: `13.0.0` is the latest version, but it requires PostCSS 8. Currently used dependencies are based on the lower versions and it looks like PostCSS 7 is used at the end:
 * https://github.com/facebookincubator/infima/blob/master/yarn.lock#L3983-L4027

kimbaudi:
Hi, I discovered this project after discovering Docusaurus v2 last week. When I first looked at the css files in this repo, I thought something about the css is a bit weird. First of all, the extension was `.css`, but it had more of a SCSS syntax except it also had some other strange syntax.  For example, I never saw carat symbol (`^`) in SCSS until I looked at the css files in this repo:

**packages\core\styles\components\dropdown.css**:

```scss
.dropdown {
  display: inline-flex;
  font-weight: var(--ifm-dropdown-font-weight);
  position: relative;
  vertical-align: top;

  &--hoverable:hover,
  &--show {
    ^&__menu {
      opacity: 1;
      transform: scale(1);
      visibility: visible;
    }
  }
  /*...*/
}
```

Later, I discovered they are using [postcss-nested-ancestors](https://github.com/toomuchdesign/postcss-nested-ancestors) plugin, which introduces `^&` selector to reference parent ancestor selectors.

Also, they use [postcss-each](https://github.com/outpunk/postcss-each) plugin instead of standard [SCSS @each](https://sass-lang.com/documentation/at-rules/control/each).

**packages\core\styles\layout\spacing.css** using PostCSS:

```scss
@each $type in (margin, padding) {
  @each $size, $value in (none, xs, sm, md, lg, xl),
    (0, 0.25rem, 0.5rem, 1rem, 2rem, 5rem)
  {
    .$(type)--$(size) {
      $(type): $(value) !important;
    }
    /*...*/
  }
}
```

**packages\core\styles\layout\spacing.scss** converted to SCSS:

```scss
@each $type in (margin, padding) {
  @each $size, $value in (none 0), (xs 0.25rem), (sm 0.5rem), (md 1rem), (lg 2rem), (xl 5rem)
  {
    .#{$type}--#{$size} {
      #{$type}: #{$value} !important;
    }
    /*...*/
  }
}
```

I'm not sure if the end goal is to convert as much of PostCSS stuff into SCSS stuff or just change file extension from `.css` to `.scss` and still keep using PostCSS plugins, but I wanted to point out that __it would require much more changes than converting file extension from `.css` to `.scss` if you are trying to remove some dependencies on PostCSS plugins and just use SCSS__. 

There are some things that might be better to convert to SCSS instead of using PostCSS:

- use standard SCSS `@each` instead of `postcss-each` in `packages\core\styles\layout\spacing.css`.
- remove using `postcss-color-mod-function` since `color-mod()` has been removed from [Color Module Level 4 specification](https://www.w3.org/TR/css-color-4/#changes-from-20160705). 

You should be able to [create tint and shade functions in SCSS](https://css-tricks.com/snippets/sass/tint-shade-functions/) and use that instead of `color-mod` in `packages\core\styles\common\variables.css`

**packages\core\styles\common\variables.css**:

```scss
/// Slightly lighten a color
/// @access public
/// @param {Color} $color - color to tint
/// @param {Number} $percentage - percentage of `$color` in returned color
/// @return {Color}
@function tint($color, $percentage) {
  @return mix(white, $color, $percentage);
}

/// Slightly darken a color
/// @access public
/// @param {Color} $color - color to shade
/// @param {Number} $percentage - percentage of `$color` in returned color
/// @return {Color}
@function shade($color, $percentage) {
  @return mix(black, $color, $percentage);
}

$ifm-color-primary: #3578e5;
$ifm-color-secondary: #ebedf0;
$ifm-color-success: #00a400;
$ifm-color-info: #54c7ec;
$ifm-color-warning: #ffba00;
$ifm-color-danger: #fa383e;

$ifm-dark-value: 10%;
$ifm-darker-value: 15%;
$ifm-darkest-value: 30%;
$ifm-light-value: 15%;
$ifm-lighter-value: 30%;
$ifm-lightest-value: 50%;

:root {
  // Colors.
  --ifm-color-primary: #{$ifm-color-primary};
  --ifm-color-primary-dark: #{shade($ifm-color-primary, $ifm-dark-value)};
  --ifm-color-primary-darker: #{shade($ifm-color-primary, $ifm-darker-value)};
  --ifm-color-primary-darkest: #{shade($ifm-color-primary, $ifm-darkest-value)};
  --ifm-color-primary-light: #{tint($ifm-color-primary, $ifm-light-value)};
  --ifm-color-primary-lighter: #{tint($ifm-color-primary, $ifm-lighter-value)};
  --ifm-color-primary-lightest: #{tint($ifm-color-primary, $ifm-lightest-value)};

  --ifm-color-secondary: #{$ifm-color-secondary};
  --ifm-color-secondary-dark: #{shade($ifm-color-secondary, $ifm-dark-value)};
  --ifm-color-secondary-darker: #{shade($ifm-color-secondary, $ifm-darker-value)};
  --ifm-color-secondary-darkest: #{shade($ifm-color-secondary, $ifm-darkest-value)};
  --ifm-color-secondary-light: #{tint($ifm-color-secondary, $ifm-light-value)};
  --ifm-color-secondary-lighter: #{tint($ifm-color-secondary, $ifm-lighter-value)};
  --ifm-color-secondary-lightest: #{tint($ifm-color-secondary, $ifm-lightest-value)};

  --ifm-color-success: #{$ifm-color-success};
  --ifm-color-success-dark: #{shade($ifm-color-success, $ifm-dark-value)};
  --ifm-color-success-darker: #{shade($ifm-color-success, $ifm-darker-value)};
  --ifm-color-success-darkest: #{shade($ifm-color-success, $ifm-darkest-value)};
  --ifm-color-success-light: #{tint($ifm-color-success, $ifm-light-value)};
  --ifm-color-success-lighter: #{tint($ifm-color-success, $ifm-lighter-value)};
  --ifm-color-success-lightest: #{tint($ifm-color-success, $ifm-lightest-value)};

  --ifm-color-info: #{$ifm-color-info};
  --ifm-color-info-dark: #{shade($ifm-color-info, $ifm-dark-value)};
  --ifm-color-info-darker: #{shade($ifm-color-info, $ifm-darker-value)};
  --ifm-color-info-darkest: #{shade($ifm-color-info, $ifm-darkest-value)};
  --ifm-color-info-light: #{tint($ifm-color-info, $ifm-light-value)};
  --ifm-color-info-lighter: #{tint($ifm-color-info, $ifm-lighter-value)};
  --ifm-color-info-lightest: #{tint($ifm-color-info, $ifm-lightest-value)};

  --ifm-color-warning: #{$ifm-color-warning};
  --ifm-color-warning-dark: #{shade($ifm-color-warning, $ifm-dark-value)};
  --ifm-color-warning-darker: #{shade($ifm-color-warning, $ifm-darker-value)};
  --ifm-color-warning-darkest: #{shade($ifm-color-warning, $ifm-darkest-value)};
  --ifm-color-warning-light: #{tint($ifm-color-warning, $ifm-light-value)};
  --ifm-color-warning-lighter: #{tint($ifm-color-warning, $ifm-lighter-value)};
  --ifm-color-warning-lightest: #{tint($ifm-color-warning, $ifm-lightest-value)};

  --ifm-color-danger: #{$ifm-color-danger};
  --ifm-color-danger-dark: #{shade($ifm-color-danger, $ifm-dark-value)};
  --ifm-color-danger-darker: #{shade($ifm-color-danger, $ifm-darker-value)};
  --ifm-color-danger-darkest: #{shade($ifm-color-danger, $ifm-darkest-value)};
  --ifm-color-danger-light: #{tint($ifm-color-danger, $ifm-light-value)};
  --ifm-color-danger-lighter: #{tint($ifm-color-danger, $ifm-lighter-value)};
  --ifm-color-danger-lightest: #{tint($ifm-color-danger, $ifm-lightest-value)};
  /*...*/
}
```

Also, want to point out that you cannot completely remove postcss because it does some really cool things that SCSS just cannot do for us. For example, [postcss-sort-media-queries](https://github.com/yunusga/postcss-sort-media-queries), which sorts `@media` queries and places them at the bottom of the processed css file.

Simek:
@yangshun

yangshun:
Thanks for the PR @Simek. The files contain some SCSS syntax now because SCSS has many of the features we need. I'd consider SCSS a small part of the CSS processing pipeline and it feels odd to rename the files to be `.scss`. The files were never meant to be SCSS files.

As the CSS spec advances, it will include more and more features which SCSS set out to solve. These would be implemented as PostCSS plugins. Hence if I had to choose a "pure" solution, I'd choose PostCSS over SCSS as it is more extensible and PostCSS IMO is the more forward-looking and powerful CSS preprocessing platform. I'd want to use a syntax which is closer to a future spec, and not a custom syntax like SCSS.

I appreciate your efforts but I'm not sure if I agree with this approach.

yangshun:
@Simek I appreciate your efforts. Looking forward to further collaboration in future. Closing this for now if there are no objections.

yangshun:
Reopening as I wrongly assumed we weren't going ahead with this. cc @Simek

Simek:
> But we are using PostCSS, so using the Sass extension is semantic wrong. Back then, `pcss` was the correct extension for style files.

Sorry but it seems you do not understand what was the goal of this PR and the architecture of pipeline. Infima is already using the SCSS syntax (https://github.com/facebookincubator/infima/blame/master/packages/postcss-preset-infima/index.js#L61) and was using it since first day of OSing. 

"Semantic wrong" is using SCSS syntax in CSS files too.

Simek:
> @Simek I appreciate your efforts. Looking forward to further collaboration in future. 

Thank you, I have left few messages and a proposition on how to move forward with this PR on Discord.

lex111:
Hmm, I'll admit I didn't really know that `postcss-scss` was being used, however it doesn't actually add SASS support as it says on [the project page](https://github.com/postcss/postcss-scss):

> **This module does not compile SCSS**. It simply parses mixins as custom at-rules & variables as properties, so that PostCSS plugins can then transform SCSS source code alongside CSS.

Thus, we can safely remove this package without any breaking changes. Try removing the line with `syntax: scss` and you will make sure yourself that. So we are not using any SASS syntax.

FYI we used double-slash comments before (`// Comment` instead of `/* Comment */`), actually this is the only thing related to SASS syntax. But now we do not use this kind of comments, so it makes no sense to use `postcss-scss`.

Simek:
> Hmm, I'll admit I didn't really know that `postcss-scss` was being used, however it doesn't actually add SASS support as it says on [the project page](https://github.com/postcss/postcss-scss):
> 
> > **This module does not compile SCSS**. It simply parses mixins as custom at-rules & variables as properties, so that PostCSS plugins can then transform SCSS source code alongside CSS.

Compile != Parse.

> Thus, we can safely remove this package without any breaking changes. Try removing the line with `syntax: scss` and you will make sure yourself that. So we are not using any SASS syntax.
>
> FYI we used double-slash comments before (`// Comment` instead of `/* Comment */`), actually this is the only thing related to SASS syntax. But now we do not use this kind of comments, so it makes no sense to use `postcss-scss`.

I did not check the whole code extensively (I'm not on FB OS payroll) and preparing this PR I assumed that if someone set syntax to SCSS in config earlier that was required for some transformation used in files. 

In this PR I was trying to improve the pipeline and match files extension to syntax (to fix highlighting issues). 

> UPD: In my view, it is worth using the `scss` extension only if purely SASS itself (not PostCSS) or its syntax is actually used. None of this is used in this project, so I honestly don't understand the meaning of this PR/discussion.

Sorry to hear that you still don't understand that `.css` is not a valid extension for PostCSS files (the final, compiled file should only have this extension, not PostCSS partials/modules). Kimbaudi also tried to explain the problem to you, I really don't know how else we can simplify this for you.

Simek:
### Preview after changes
Screenshot 2021-02-15 193903

Simek:
### LiveReload fix proof
lr

Simek:
@yangshun In the latest commit I have added to the Demo an ability to toggle between regular style and the minified version. Demo so far was using unminified version of files, when in production environment most of Infima users leverage the benefits of minified files. To keep both files testable I have introduced the toggler. 

This feature should also help developers to spot some minification pipeline issues.

To implement this properly I have also add styles minification to the `watch` task (which is used during `yarn start`). This slow down the Live Reload speed a bit, but the overhead is tolerable in my opinion.

### Preview
min

Simek:
In latest commit I have added the RTL toggler which enables to preview newly added RTL styles. 

Unfortunately RTL styles doesn't look correctly right now, it looks like more than RTLCSS conversion will be needed to ship this feature.

Simek:
@yangshun FYI: Due to #69 merge not all comments and commit messages are accurate anymore.

yangshun:
Thanks @Simek. I'm afraid I can only find time to do review this over the weekend. Larger PRs will take longer to review.

slorber:
Updated the PR to trigger the new build size bot + also fix yarn lock

slorber:
To make the review of such PR easier, I've added:
- the build size bot
- dist to git so that we can compare dist output before/after


What I understand from this PR:
- using `.pcss` extension is more correct than `.css` and improves IDE support
- it does not have any significant impact on the final `/dist` folder (apart from a removed comment)

So, if you have anything against using a `.pcss` extension I'm going to merge this PR soon.

I'll just review the other things of this PR before merging.

slorber:
LTR toggle is now working in the demo:
https://deploy-preview-58--infima.netlify.app/demo/

image

packages/core/demo/index.html
```diff
@@ -2625,23 +2637,81 @@ <h4 class="footer__title">More</h4>
...
         function toggleMode() {
-          const isDarkTheme =
-            document.querySelector('html').getAttribute('data-theme') ===
-            'dark';
-          document
-            .querySelector('html')
-            .setAttribute('data-theme', isDarkTheme ? '' : 'dark');
+          const isDarkTheme = html.getAttribute('data-theme') === 'dark';
+          html.setAttribute('data-theme', isDarkTheme ? '' : 'dark');
           history.pushState({}, null, isDarkTheme ? '/' : '/?dark=true');
```
Simek, line 2647:
```
history.pushState({}, null, isDarkTheme ? '/demo/' : '/demo/?dark=true');
```

yangshun (approved):
@Simek I think this looks great. I like this new `.pcss` extension. I guess we need to tell people to install pcss extensions and recommend them some. Few things I need you to do before I can merge this:

- Can you rebase and resolve the merge conflicts? We just added Prettier
- Would it make sense for the changes to `index.html` to be within another PR? I think we're overloading this PR with too many new features (don't get me wrong, I like them all). I'm fine if you think we can keep it within this PR

packages/core/demo/index.html
Simek: @slorber The current implementation break the demo:
bug

Simek:
> @Simek I think this looks great. I like this new `.pcss` extension. I guess we need to tell people to install pcss extensions and recommend them some. Few things I need you to do before I can merge this:
> 
> * Can you rebase and resolve the merge conflicts? We just added Prettier
> * Would it make sense for the changes to `index.html` to be within another PR? I think we're overloading this PR with too many new features (don't get me wrong, I like them all). I'm fine if you think we can keep it within this PR

@yangshun I was hoping that PR can be accepted before the Prettier changes, but I will update the PR.

Was expressing this several times already, let mi write it again. This is not the usual way I'm partitioning the PRs, usually it's one feature/bugfix per PR (#56, #57), but some improvements to the demo required small tweaks to the already changed in this PR Gulp pipeline and I feel like pushing few different PRs including some duplicate changes would only increase, the already presented confusion around this PR. 

So definitely, in the future, you can expect way smaller and better divided PRs from me.

yangshun:
> improvements to the demo required small tweaks to the already changed in this PR Gulp pipeline and I feel like pushing few different PRs including some duplicate changes would only increase, the already presented confusion around this PR.

This is where stacked PRs come in handy. Sadly GitHub doesn't support this feature

> So definitely, in the future, you can expect way smaller and better divided PRs from me.

Awesome <3

slorber:
Fixed the demo code so that querystring works reliably for all 3 toggles + with or without baseurl (as Netlify preview use baseurl)

Going to merge now!

PR #61, opened 2021-02-01
feat: add RTL support

lex111:
Adding generation of CSS bundle with styles for RTL direction. Thus, in Docusaurus core we will dynamically include the proper file (`default.css` or `default-rtl.css`), depending on the direction specified in the config file.

Of course, we can put CSS rulesets for RTL in one same default CSS file, but this will increase this file size, so it makes more sense to generate a separate bundle for this purpose.

image


slorber:
LGTM, let's try this! The

lex111:
But what about merging this PR? https://github.com/facebookincubator/infima/pull/62

PR #63, opened 2021-02-14
refactor: improve UX of mobile navbar

lex111:
Attempts to make a smoother transition of the navbar sidebar display on mobiles. Based on the mobile Wikipedia website experience.

| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (15) | ezgif com-gif-maker (16) |

Builds:

- Before: https://infima63-before.netlify.app/
- After: https://infima63-after.netlify.app/

(Experiment) new changes with using `translate3d` instead of `translateX` for better graphics performance  https://infima63-after-translate3d.netlify.app/ 

cc @yangshun what do you think about this improvement (and also to replace `translateX` with `translate3d`)?

### DIFF
--- a/packages/core/styles/components/navbar.css
+++ b/packages/core/styles/components/navbar.css
@@ -189,50 +189,40 @@
   }
 
   &-sidebar {
-    background-color: var(--ifm-navbar-background-color);
-    bottom: 0;
-    box-shadow: var(--ifm-global-shadow-md);
-    display: none;
-    left: calc(var(--ifm-navbar-sidebar-width) * -1);
     position: fixed;
     top: 0;
-    transition: transform var(--ifm-transition-fast) ease-in-out;
+    left: 0;
+    bottom: 0;
+    visibility: hidden;
+    opacity: 0;
     width: var(--ifm-navbar-sidebar-width);
+    background-color: var(--ifm-navbar-background-color);
+    box-shadow: var(--ifm-global-shadow-md);
     overflow: auto;
+    transform: translateX(-100%);
+    @mixin transition opacity visibility transform, 250ms, ease-in-out;
 
     &--show {
-      ^& {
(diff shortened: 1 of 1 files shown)


packages/core/styles/components/navbar.css
```diff
@@ -189,40 +189,40 @@
   }
 
   &-sidebar {
-    background-color: var(--ifm-navbar-background-color);
```
yangshun, line 192: Can we leave the properties sorted in alphabetical order please? This is a FB convention and I'd like to keep it

yangshun (approved):
This certainly looks nicer! I prefer the properties to be sorted alphabetically. Could you revert the order change?

lex111:
Ah, sure, fixed it! What do you think of `translate3d`? Should we use it? This should speed up performance as this navbar is for mobiles.

yangshun:
> Ah, sure, fixed it!

Thank you!

> What do you think of `translate3d`?

Yep I think it's good, thanks for using that instead!

lex111:
> Yep I think it's good, thanks for using that instead!

But I didn't actually do it in this PR! I just suggested it and built framework with proposed change. So I opened new PR for that. https://github.com/facebookincubator/infima/pull/65

packages/core/styles/components/navbar.css
```diff
@@ -192,37 +192,37 @@
...
       background-color: rgba(0, 0, 0, 0.6);
       bottom: 0;
-      display: none;
       left: 0;
-      right: 0;
+      opacity: 0;
       position: fixed;
+      right: 0;
       top: 0;
+      transition: opacity 100ms ease-in-out, visibility 0ms linear 100ms;
```
Simek, line 224: `0` values do not need an unit.

PR #64, opened 2021-02-14
fix: use button instead of div for navbar toggle button

lex111:
For improve accessibility purposes.

### DIFF
--- a/packages/core/styles/components/navbar.css
+++ b/packages/core/styles/components/navbar.css
@@ -83,6 +83,9 @@
     cursor: pointer;
     display: none;
     margin-right: 0.5rem;
+    border: 0;
+    background: 0 0;
+    padding: 0;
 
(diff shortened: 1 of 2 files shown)


yangshun (approved):
Do we need to change this on Docusaurus classic theme side?

lex111:
Yes, of course I will do it after the new release of Infima.

packages/core/styles/components/navbar.css
```diff
@@ -83,6 +83,9 @@
     cursor: pointer;
     display: none;
     margin-right: 0.5rem;
+    border: 0;
+    background: 0 0;
```
Simek, line 87: If the goal here was to reset the background position the background-position property should be used.

If the goal was to remove background completely using none is recommended over 0 and the second 0 is redundant.

PR #66, opened 2021-02-14
refactor: sort CSS properties alphabetically

lex111:
I thought that we need to use stylelint to enforce the code style. This PR contains the minimum setup for that. Mainly to sort the CSS properties in alphabetical order.

But it's better of course better to automate this using GitHub actions and husky/lint-staged (locally, pre-commit hook).

Also it seems we need to choose a config (https://github.com/stylelint/awesome-stylelint#configs), @yangshun could you please tell me the config we need? Which one is commonly used in Facebook?


packages/core/styles/content/code.css
```diff
@@ -48,7 +48,6 @@ pre {
     display: inline;
     font-size: 100%;
     line-height: inherit;
-    overflow: visible;
```
lex111, line 51: Not need (fixed by https://github.com/kristerkari/stylelint-declaration-block-no-ignored-properties)

packages/core/styles/components/navbar.css
```diff
@@ -24,7 +24,6 @@
 }
 
 .navbar {
-  // TODO: Add navbar-link-active
```
lex111, line 27: It's already done

yangshun (approved):
LGTM. The FB stylelint rules are not public, so we can't reuse them. But maybe we could use the stylelint-copyright rule in https://github.com/facebook/docusaurus/blob/master/.stylelintrc.js

Simek:
@yangshun This PR includes the stealth conversion of the inline comments to regular ones, which make the [`postcss-strip-inline-comments`](https://github.com/mummybot/postcss-strip-inline-comments) plugin redundant (https://github.com/facebookincubator/infima/blob/master/packages/postcss-preset-infima/index.js#L15) and additionally `dist` files are containing all the comments from the partials now.

PR #67, opened 2021-02-15
fix: improve stylelint

lex111:
- integrate of lint-staged/husky
- add stylelint-copyright check
- add stylelint-high-performance-animation check and fix bugs that it reported (use `pointer-events` instead of `visibility`)

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,19 +29,24 @@
     "gulp-rtlcss": "^1.4.1",
     "gulp-sass": "^4.0.1",
     "gulp-webserver": "^0.9.1",
+    "husky": "4",
+    "lint-staged": "^10.5.4",
     "natives": "^1.1.6",
     "postcss-preset-infima": "^0.2.0-alpha.18",
     "stylelint": "^13.10.0",
+    "stylelint-copyright": "^2.0.0-alpha.69",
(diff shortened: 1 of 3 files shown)


slorber:
LGTM 👍

packages/core/package.json
```diff
@@ -29,33 +29,52 @@
...
   "scripts": {
-    "start": "gulp",
     "build": "gulp build",
+    "start": "gulp",
+    "lint": "stylelint \"styles/**/*.css\"",
     "lint:fix": "stylelint \"styles/**/*.css\" --fix"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
```
Simek, line 49: You have bumped the Node version in the core package, but you have forgot to update the Node in Infima PostCSS preset `package.json`.

packages/core/package.json
```diff
@@ -29,33 +29,52 @@
     "gulp-rtlcss": "^1.4.1",
     "gulp-sass": "^4.0.1",
     "gulp-webserver": "^0.9.1",
+    "husky": "4",
```
Simek, line 32: What is the reason behind specifying only major version instead of full one (latest: `4.3.8`) and omitting the caret?

PR #70, opened 2021-02-18
chore: monitor dist size changes in PRs

slorber:
We should be able to detect easily when the dist output is affected by a PR.

Asked by @Simek in https://github.com/facebookincubator/infima/pull/58#issuecomment-737512090

> Also as I was suggesting to @yangshun before, it might be nice to setup the dist CSS diff workflow on CI, so we can be sure that the output before and after the changes will be the same.

Is this way you had in mind or something more advanced?

### DIFF
--- a/.github/workflows/build-size.yml
+++ b/.github/workflows/build-size.yml
@@ -1,0 +1,17 @@
+name: Build Size Report
+
+on:
+  pull_request_target:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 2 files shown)


yangshun (approved):
Whoa this is awesome. Didn't know it could be done so easily

slorber:
actually I think @Simek set it up initially on docusaurus :)

PR #73, opened 2021-02-18
PR to test build size bot 2

slorber:


slorber:
So, it seems there's a bug in the build size bot.

Will keep it for now

PR #74, opened 2021-02-18
chore: build-size-bot fix

slorber:
The bot does not support "pull_request_target" apparently (which we need to post a PR comment on external contributors PRs)

https://github.com/preactjs/compressed-size-action/issues/54

Until it is fixed, we'll at least have the build size report for internal contributors 😅


LongLiveKS3 (commented):
Ok

PR #82, opened 2021-02-19
chore: Update Netlify to publish site + demo

slorber:
We want to deploy the infima Docusaurus site in deploy previews so that we can detect regressions for a Docusaurus site using the classic theme.

Also deploys the existing HTML demo, but we'd rather migrate it as a standalone page of the site?

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -5,75 +5,88 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+// See Netlify env variables here: https://docs.netlify.com/configure-builds/environment-variables/#build-metadata
+const isNetlify = process.env.NETLIFY === "true";
+const isNetlifyDeployPreview =
+  isNetlify && process.env.CONTEXT === "deploy-preview";
+
 module.exports = {
-  title: 'Infima',
-  tagline: 'A modern styling framework for content-driven websites 🔥',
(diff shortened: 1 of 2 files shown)


website/docusaurus.config.js
```diff
@@ -5,75 +5,88 @@
...
+
 module.exports = {
-  title: 'Infima',
-  tagline: 'A modern styling framework for content-driven websites 🔥',
-  organizationName: 'facebookincubator',
-  projectName: 'infima',
-  baseUrl: '/infima/',
-  url: 'https://facebookincubator.github.io',
-  favicon: 'img/logo.png',
+  title: "Infima",
```
yangshun, line 14: Hmm why are all these quotes changed? Do you have a different Prettier settings?

yangshun (approved):
LGTM except for the redundant styling changes in `docusaurus.config.js`

website/docusaurus.config.js
yangshun: There's no Prettier config in the root of the repo. I'll add one.

slorber: yes, actually. there is none and planned to add it next :)
Found it was not too bad to add these changes in the meantime

yangshun: Here it is - https://github.com/facebookincubator/infima/pull/83

PR #83, opened 2021-02-19
feat: add prettier command

yangshun:
Prettify js and MDX files and add lint-staged commands for them


slorber:
@yangshun I was wondering if it's possible to prettify .pcss files too?

Never used prettier for that and my first attempt was not very successful

yangshun:
I'll look into doing that

Issue #85, opened 2021-02-19
Upgrade to Node 12


slorber:
Docusaurus moved to Node 12, as Node 10 EOL is soon.

Infima requires Node 10 to build correctly and fails on Node 12. 

This is annoying because we can't use latest Docusaurus on Infima website (including deploy previews).

We should make sure Infima is able to build on Node 12 or we won't be able to upgrade the Infima website to latest Docusaurus.

```
yarn install v1.22.5
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error @docusaurus/core@2.0.0-alpha.094b51390: The engine "node" is incompatible with this module. Expected version ">=12.13.0". Got "10.23.3"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

PR #86, opened 2021-02-19
chore: update PostCSS to v8, bump Node version and other dependencies

Simek:
This PR bumps PostCSS to version `8` (changes performed according to the migration guide - https://github.com/postcss/postcss/wiki/PostCSS-8-for-end-users), minimum Node version to `12` (as requested in #85) and other PostCSS related plugins which cannot be updated earlier due to pin to PostCSS `7`.

Those change should not affect the output files at all, but let's run the CI and see the results.

CC: @slorber


slorber:
great thanks, that seems to have no impact :)

PR #87, opened 2021-02-19
misc: wire lint-staged + add lint CI

slorber:
The prettier lint-staged was not run as it was not wired to git precommit.

Also add a GH CI lint check because devs can still commit and disable the pre-commit hooks locally


packages/core/styles/content/code.pcss
```diff
@@ -18,6 +17,9 @@
   --ifm-pre-color: inherit;
   --ifm-pre-line-height: 1.45;
   --ifm-pre-padding: 1rem;
+
+
+  --ifm-code-font-size: 95%;
```
yangshun, line 22: We'll revert this right?

slorber: yes I'm testing commiting things and disabling the precommit hook :) as expected the CI fails because /dist was not updated

PR #88, opened 2021-02-19
chore: add yarn-deduplicate to the postinstall, clean lock

Simek:
As seen in Docusuaurs 2 repo (https://github.com/facebook/docusaurus/pull/3738), besides initial concerns, adding [`npx yarn-deduplicate`](https://github.com/atlassian/yarn-deduplicate) isn't a bad idea and can be used as a long term solution for keeping `yarn.lock` tidy.

This PR adds the deduplication script to the `postinstall` of the workspace and includes the changes added by running the script.

CC @slorber


slorber:
thanks 👍

PR #94, opened 2021-03-09
release: v0.2.0-alpha.20

slorber:

PR #96, opened 2021-03-13
chore: update publish guide

Simek:
Update publish guide, to minimize the risk of misaligned dependencies in the future.

Refs #94, #95


slorber:
Thanks, didn't even know this was needed 😅

PR #101, opened 2021-03-16
chore: update publishing guide again

Simek:
Refs #98 

Add one more step to ensure that lockfile is regenerated properly after the packages version bump.


slorber:
thanks 👍

PR #106, opened 2021-04-10
fix: make sidebar menu arrow position fixed

lex111:
Should fix https://github.com/facebook/docusaurus/issues/4565

| Before   | After    |
| -------- | -------- |
| image | image |


yangshun (approved):

yangshun:
@lex111 could you resolve the conflicts and self-merge (forgot if you have permissions).

slorber:
@Simek reported that there is a new border on code items in the TOC, is this on purpose?

https://reactnative.dev/docs/scrollview
VS
https://deploy-preview-2573--react-native.netlify.app/docs/scrollview

image


This border is almost invisible in the dark theme BTW.

I don't have a strong feeling about adding a border or not in this case, but maybe it is preferable to keep former behavior (no border) as nobody complained about it?

lex111:
Border is needed, this one works fine with default styles (as
intended). Of course, in the case of large customizations, it may not
fit, but then you can just manually remove it in custom CSS file,
right?

2021-05-06 11:45 GMT+03:00, Sébastien Lorber ***@***.***>:

Simek:
> Border is needed

Can you explain why?

lex111:
To better separate background color of inline code from foreground of page. This is common design pattern, so nothing unusual here.

Simek:
> To better separate background color of inline code from foreground of page. This is common design pattern, so nothing unusual here.

So it's not needed, (it's only a visual improvement in some cases). Also I wonder why this UI change has been smuggled in non-related sidebar fix?

slorber:
There is a PR about inline code blocks.
I suggest we move the discussions on the appropriate PR:   https://github.com/facebookincubator/infima/pull/113

PR #107, opened 2021-04-11
Truncate long navbar title with ellipses

lex111:
Several times I've come across Docusaurus sites that have a long enough navbar title which looks bad on mobiles. I suggest truncating it with ellipsis. It will take some code to achieve this, but I think it makes sense to make the design more responsive.

On [keydb](https://docs.keydb.dev/docs/migration/) website as example:

| Before   | After    |
| -------- | -------- |
| image | image |

| Before   | After    |
| -------- | -------- |
| image | image |

| Before   | After    |
| -------- | -------- |
| image | image |


yangshun (approved):

benschermel:
We have updated this fix on our website (KeyDB) and it looks way better!, thanks @lex111. You can see how it looks [here](https://docs.keydb.dev/docs)

slorber:
thanks for the feedback @benschermel 

don't forget to add your site to our showcase ;)

https://docusaurus.io/showcase

PR #112, opened 2021-04-24
fix: adjust dropdown position

lex111:
Related to https://github.com/facebook/docusaurus/issues/4671

Make vertical center alignment of dropdown icon and padding between dropdown items completely identical as earlier.


yangshun (approved):
Thanks, do you know what caused this?

lex111:
I think, technically this icon is vertically centered correctly, but visually gives the impression that it is slightly higher positioned relative to the link text.

PR #113, opened 2021-04-24
refactor: restyle inline code

lex111:
An attempt to make inline code elements neater and less contrasting. Also resolves the issue where the inline code is not visible in the secondary admonition (since the colors are identical).

## Before

Admonition (secondary):

| Light mode   | Dark mode    |
| -------- | -------- |
| image  | image |

Text: 

| Light mode   | Dark mode    |
| -------- | -------- |
| image | image |

## After

Admonition (secondary):

| Light mode   | Dark mode    |
| -------- | -------- |
| image  | image  |

Text: 

| Light mode   | Dark mode    |
| -------- | -------- |
|  image |  image |


packages/core/styles/content/code.pcss
```diff
@@ -41,17 +45,13 @@ pre {
...
-  word-wrap: normal;
 
   code {
     background-color: transparent;
-    display: inline;
+    border: none;
     font-size: 100%;
     line-height: inherit;
     padding: 0;
-    white-space: pre;
```
lex111, line 52: Since these are default values, we can safely remove them.

yangshun (approved):

slorber:
Reposting comment of wrong PR:

---

@Simek reported that there is a new border on code items in the TOC, is this on purpose?

https://reactnative.dev/docs/scrollview
VS
https://deploy-preview-2573--react-native.netlify.app/docs/scrollview

image


This border is almost invisible in the dark theme BTW.

I don't have a strong feeling about adding a border or not in this case, but maybe it is preferable to keep former behavior (no border) as nobody complained about it?

slorber:
Looks like there's a regression when using link + inline code:

https://github.com/facebook/docusaurus/issues/4781

image

lex111:
No, it's not related to this PR, but I tried it fixed in my new PR https://github.com/facebookincubator/infima/pull/120

PR #116, opened 2021-04-24
refactor: some cleanups

lex111:
Removing unnecessary, meaningless properties and transform some of them into shorthand form.


yangshun (changes requested):
We want to be explicit and avoid shorthand where possible (coding style decision). Also because we use alphabetical ordering of the properties, sometimes shorthand can be confusing.

lex111:
Actually longhand properties create difficulties for understanding, here is a quick example:

image

Due to alphabetical order of properties, the `boder-radius` property sits between the `border`-properties, which is really confusing.

But I removed such shorthand form so this PR can be merged.

PR #117, opened 2021-04-24
refactor: add double bottom border to thead

lex111:
In many CSS frameworks, the table header is visually distinctly separated from the regular rows, so I suggest follow such pattern in Infima.

| Before   | After    |
| -------- | -------- |
| image | image |


yangshun (approved):
Let's start using explicit properties instead of shorthand.

lex111:
Why did you change coding style? Doesn't longhand properties make the code less readable and increase the bundle in the end? I've rarely seen `border` property written in longhand form, it seems like excessive. Is not it so?

lex111:
Although such a code decision does seem very counterproductive, I replaced shorthand form with longhand one.

PR #120, opened 2021-05-14
refactor: use color with opacity for inline code

lex111:
We're probably better off using colors with opacity for inline code. A more generic approach that is suitable for all use cases. 

Before:

| Light   | Dark    |
| -------- | -------- |
| image | image |

After:

| Light   | Dark    |
| -------- | -------- |
| image | image |


slorber:
looks better, any opinion @yangshun ?

slorber (approved):

yangshun (approved):

PR #121, opened 2021-05-16
refactor: use relative positioning wherever possible

lex111:
Absolute positioning often requires more code, so it makes sense to get rid of it in favor of relative positioning (if possible).


lex111:
This will also remove extra padding-right on a external link.

image

lex111:
I'm going to merge this, since it seems like safe changes to me. Especially, this is the last open PR, so I can do a new release after merging it.

PR #125, opened 2021-05-21
fix: remove padding-left from last navbar item

lex111:
| Before   | After    |
| -------- | -------- |
| image | image |

@yangshun I'm actually wondering why `padding` is used for navrbar item links rather than regular `margin-left`? It probably isn't very useful to increase the clickable area of a navbar link in this way. Maybe we should replace `padding` with `margin`? (Then, we don't need to use new `last-child` rule.)


yangshun (changes requested):
It's intentional, all the nav items have some padding around it to increase the clickable area,

lex111:
Alright, but for the last navbar item can we remove `padding-left` as proposed in this PR?

packages/core/styles/components/navbar.pcss
```diff
@@ -122,6 +122,10 @@
       flex: 0 0 auto;
       justify-content: flex-end;
     }
+
+    > :last-child {
```
yangshun, line 126: Maybe this should only be for items in the `--right` side

yangshun (approved):
Left a comment.

packages/core/styles/components/navbar.pcss
lex111: Agree, will do it. BTW what do you think about this PR #121?

PR #128, opened 2021-06-01
fix: make dropdown menu scrollable

lex111:
If there are a lot of elements in dropdown menu on site, then not all of them will be visible in the viewport. Currently, this problem can be seen on site https://relay.dev/docs/.

| Before   | After    |
| -------- | -------- |
|  image |  image |


slorber:
Make sense, thanks

Probably won't work if user want to use a dropdown inside the content though (I don't think Infima assume the dropdown will always be in the navbar), but this looks like a good-enough solution for 99% of use-cases and improves from the current state.

PR #129, opened 2021-06-01
refactor: make dropdown menu transition less jumpy

lex111:
| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (2) | ezgif com-gif-maker (3) |


slorber (approved):
No strong opinion about this one, both looks good to me

lex111:
In my opinion, new transition is less "aggressive", smoother and more pleasant.

PR #130, opened 2021-06-01
refactor: apply only margin top to menu list item

lex111:
A more accurate way to set margin for menu list items to avoid unnecessary margin on last item.


slorber:
Not against, but comparing before/after it does not seem to make any visual difference? not sure why

Comparing
- https://infima.dev/demo/
- https://deploy-preview-130--infima.netlify.app/demo/

There seems to be some kind of margin collapsing because applying `margin-bottom: 0.25rem;` or `margin-bottom: 0;` does not change anything to me, but `margin-bottom: 0.5rem;` does change something

lex111:
The difference will be visible in "real" sidebar, for example on Docusaurus:

| Before   | After    |
| -------- | -------- |
| image | image |

slorber (approved):

PR #132, opened 2021-06-06
feat: add utility class for reset button

lex111:
In Docusaurus, we currently have 5 custom buttons, in which we must first reset its default styles before using. I suggest introducing a helper class to facilitate this task, then in the Docusaurus codebase we can get rid of unnecessary CSS properties, so I think it makes sense.


lex111:
@yangshun you don't mind adding this new utility class?

lex111:
@slorber what do you think about this PR?

slorber:
looks fine, any opinion @yangshun ?

slorber (approved):

PR #136, opened 2021-06-10
refactor: adjust background color of dark navbar

lex111:
Related with https://github.com/facebook/docusaurus/issues/3937

Dark style of navbar should be the same color as "default" navbar when dark mode is enabled. This is the most expected and sensible behavior if the user only wants to have dark navbar  whose background color should be identical to that of the navbar in dark mode.

Before:

image

After:

image


slorber:
I don't understand the explainations 😅 After looks fine though 👍

lex111:
If you enable dark mode on Docusaurus website, you will notice that the navbar background has changed from white to dark color (which is expected). So I suggest using that color for _dark style_ of navbar.

slorber (approved):
Ah I see 👍 

Just wondering why we need to redeclare that color again, can't we share the CSS selector for dark theme navbar and dark navbar?

lex111:
It doesn't work that way, because the background color of the navbar relies on the `ifm-background-surface-color` variable, which changes depending on selected color mode, meanwhile for the dark navbar we need unchanged (dark) color.

PR #137, opened 2021-06-11
refactor: reduce padding between navbar items

lex111:
Currently, the space (padding) between navbar links is 32px -- that's quite a lot, so I suggest lowering it a bit to 24px. For example, default templates in static docs generators like VuePress or Docute use roughly the equivalent margin value for navbar items.

Given that we have already made it possible to truncate navbar title, reducing space between navbar items will allow us to fix (or at least mitigate) this issue https://github.com/facebook/docusaurus/issues/3820 (in most cases I think it will be enough, I guess).

Before:

image

After:

image


slorber:
LGTM, agree we could use a bit less spacing

PR #138, opened 2021-06-11
fix: increase secondary content color contrast to pass WCAG AAA

lex111:
Related to https://github.com/facebook/docusaurus/issues/4947

Currently, secondary content color is used for links in TOC of page and sublabel of pagination nav.

Link to checker of changed color: https://webaim.org/resources/contrastchecker/?fcolor=525860&bcolor=FFFFFF

Before:

image

After:

image


slorber:
For some reasons I don't see this change being used anywhere in a Docusaurus site. Your screenshot show a TOC but it didn't change the TOC color for me on the preview?

Otherwise 👍  for increasing contrast

slorber:
I see now. Is this a change for Infima only, or it will affect Docusaurus as well (and how?). Do you plan to use that css var in Docusaurus more?

lex111:
Since Docusaurus renders TOC, so the color will change there too. Don’t think this variable will be used often, so let's just change the color for better a11y.

slorber (approved):

PR #140, opened 2021-06-17
fix: proper styling for dark style of navbar

lex111:
Related to https://github.com/facebook/docusaurus/issues/3937

Resolving this required placing utility classes in front of the components so that the components could override the properties set by the utility classes.


packages/core/dist/css/default-dark/default-dark-rtl.css
```diff
@@ -2425,13 +2569,19 @@ hr:after {
 
 .navbar--dark {
     --ifm-navbar-background-color: #242526;
-    --ifm-navbar-link-color: var(--ifm-color-white);
+    --ifm-navbar-link-color: var(--ifm-color-gray-100);
```
lex111, line 2572: The same color is used in dark mode, so I changed it here for consistency.

packages/core/styles/components/navbar.pcss
```diff
@@ -23,6 +23,12 @@
   --ifm-navbar-sidebar-width: 80vw;
 }
 
+html[data-theme='dark'],
+.navbar--dark {
+  --ifm-menu-link-sublist-icon-filter: invert(100%) sepia(94%) saturate(17%)
```
lex111, line 28: Now `ifm-menu-link-sublist-icon-filter` var is redefined directly in the navbar file, so as not to duplicate the value unnecessarily.

slorber:
LGTM 👍

PR #141, opened 2021-06-22
fix: apply --ifm-heading-font-family to headings

dsmmcken:
Root variable was created but never actually assigned to the headers.

### DIFF
--- a/packages/core/styles/content/heading.pcss
+++ b/packages/core/styles/content/heading.pcss
@@ -28,6 +28,7 @@ h4,
 h5,
 h6 {
   color: var(--ifm-heading-color);
+  font-family: var(--ifm-heading-font-family);
   font-weight: var(--ifm-heading-font-weight);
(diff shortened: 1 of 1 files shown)


lex111 (commented):
This is unnecessary because heading font is implicitly inherited from root element `html`, so let's remove that variable.

dsmmcken:
It's extremely common to use a separate font-stack for headings vs body. This should be supported, Facebook themselves use "Optimistic Display" for headings and "Optimistic Text" for body on most corporate sites.

That's how I stumbled across this:
1. Went to set headings to a different brand related font for my site
2. Saw variable available to do so already on root
3. Set variable, saw no effect 😞 
4. Found variable is not applied to anything

lex111:
I see, I'm fine with it, going to merge this.

lex111 (approved):

lex111:
@dsmmcken thanks!

PR #146, opened 2021-07-07
refactor: make scrollable sidebar menu only

lex111:
It make sense made only the sidebar menu area scrollable, instead of _entire_ sidebar. This seems like more obvious behavior than scrolling through the whole sidebar content.

For example, this behavior is implemented in Material for MkDocs (demo - https://fastapi.tiangolo.com/)

ezgif com-gif-maker (5)

* * *

ezgif com-gif-maker (4)

It is compatible with new mobile UX:

ezgif com-gif-maker (6)


packages/core/styles/components/navbar.pcss
```diff
@@ -259,8 +259,15 @@ html[data-theme='dark'],
         var(--ifm-navbar-padding-horizontal);
     }
 
-    &__items > div {
-      padding: 0.5rem;
+    &__items {
+      display: flex;
```
slorber, line 263: is the flex prop required if the height is already explicitly set?

slorber (approved):
👍 looks better

wonder if we can keep the back button accessible when we scroll?

packages/core/styles/components/navbar.pcss
lex111: Yes, this will only make this block (items) scrollable.

lex111:
> wonder if we can keep the back button accessible when we scroll?

What do you have in mind? Could you show with example?

slorber:
> What do you have in mind? Could you show with example?

I mean only the part under the back button would scroll, but the back-button is not part of the header because it remains part of the animated pane slide transition

slorber:
image

lex111:
Since this button is part of sidebar item (menu), it will also be scrollable, suppose that would be okay.

slorber:
If the menu is scrolled down, then the back button is not accessible anymore without the user to scroll up, which may not be very convenient.

It's not implemented but I'd like the current doc item of the sidebar to `el.scrollIntoView` on navigation, so that on mobile sidebar opening the current doc is always visible (it may not, particularly when using the next button). In such case the back button would simply be hidden from the very start, so I think having a sticky button makes sense.

Anyway the current improvement is already nice so we can merge it now if we agree on it

PR #148, opened 2021-07-08
refactor: use Chrome colors for custom thin scrollbar

lex111:
Relevant for Linux/Windows (colors matched), not sure about macOS.

| Light mode   | Dark mode    |
| -------- | -------- |
| image | image |


lex111:
Maybe we should remove `border-radius` from thin scrollbar? At least on Linux/Windows it is not used for native scrollbar.

slorber:
Not exactly the same on macos, but it looks close enough for me, so let's move on

image

image

> Maybe we should remove border-radius from thin scrollbar? At least on Linux/Windows it is not used for native scrollbar.

I think it's better with the border-radius for secondary scrollbars

PR #149, opened 2021-07-08
refactor: avoid top margin for first menu item

lex111:
Follow-up of #130

A more correct solution would be to exclude top margin for first menu items.


slorber:
👍

PR #150, opened 2021-07-08
refactor: remove redundant CSS declarations

lex111:
It's unnecessary code from which we can easily remove, and nothing breaks.


packages/core/styles/components/close.pcss
```diff
@@ -6,10 +6,7 @@
  */
 
 .close {
```
slorber, line 8: is this close class still used anywhere?

slorber (approved):

packages/core/styles/components/close.pcss
lex111: No, `close` class used only in context of alerts.

slorber: I see, so it's only used in Infima and not Docusaurus? 
So should we remove this?

Related the other issue, and in my opinion, we should only keep in Infima what we use in Docusaurus theme classic, but it really depends on what we want Infima to be in the end.

@yangshun what do you think?

lex111: This class is used by the PWA plugin:

image

slorber: I see thanks 👍 let's keep it then 🤪

PR #151, opened 2021-07-08
refactor: add new mobile menu

lex111:
Since we are introducing new mobile UX in https://github.com/facebook/docusaurus/pull/4273, we can already remove styles for old-responsive sidebar menu.

@slorber it's just not clear to me -- do we need to implement secondary menu functionality in Infima itself as before in `menu.js` file?


slorber (approved):

slorber:
> @slorber it's just not clear to me -- do we need to implement secondary menu functionality in Infima itself as before in menu.js file?

My position about Docusaurus<->Infima is that both are quite coupled and I don't think many users will use Infima outside Docusaurus users anyway. 

I was in favor of moving it to the Docusaurus repo and making it an implementation detail of the classic theme, not necessarily recommend the usage of it for users, so that they could easily switch from one theme to another without re-implementing their pages.

You and @yangshun didn't really agree with this idea and wanted to keep it as a separate repo/project, so I don't have a strong opinion about which CSS rule should be in Infima or Docusaurus 😅 if it was me I'd just put everything in the Docusaurus repo and even rename Infima as `docusaurus-theme-classic-css` or something.

Now about the JS question, not 100% sure but it actually looks unused on Docusaurus, so it basically only has value for the demo page and the users of Infima outside Docusaurus. Not sure it's something we should invest in, as Docusaurus is unlikely to benefit from this.

I still see value in using some vanilla JS in Docusaurus: it would permit to mark some JS items as critical, and make them interactive before React hydration. But it's an effort that is not started yet and I'm not even sure it's worth investing in this, as React selective hydration and server components might make this optimization less useful.

lex111:
If we don't implement the secondary menu in Infima, then it doesn't make sense to include some styles for that, right?

For example, the code below is essentially needed for the secondary menu to work. And if this menu is not implemented in Infima itself, then there's not much point in related styles in the final bundle.

https://github.com/facebookincubator/infima/blob/0d50f443abdad85cbac4a49cb175c33ca8021b9c/packages/core/styles/components/navbar.pcss#L266-L270

Therefore, will implement secondary menu on Infima's demo page to follow the previously chosen approach. It also helps to keep secondary menu styles in one place (Infima), and in Docusaurus code we don't need to write any extra code for that. WDYT?

slorber:
In practice any CSS added to Infima is some additional dev friction for me so I'd rather have everything in Docusaurus. 

If you really want to have more CSS in Infima, it's up to you and @yangshun, and I'll follow what you decide but it's not the choice I would make.

lex111:
I just like the consistency, so I think it makes sense to stick with the previous approach.

slorber (approved):
That looks fine to me.
Just afraid it would add dev friction it we later need to do additional changes.
For example I think we could later replace the back arrow with a proper svg or something, and this would probably require changing Infima again. I'm not too back of these back-and-forth due to having 2 repos instead of 1.

If you think we should merge it, feel free to do so and cleanup on Docusaurus side

lex111:
> Just afraid it would add dev friction it we later need to do additional changes.

Do you mean that we will need to make changes to appropriate React component in Docusaurus? Yes, of course, will prepare PR for this.

> For example I think we could later replace the back arrow with a proper svg or something, and this would probably require changing Infima again. I'm not too fan of these back-and-forth due to having 2 repos instead of 1.

This can be done without changing Infima code, here we just provide basic styles to make two menus work.

slorber:
> Do you mean that we will need to make changes to appropriate React component in Docusaurus? Yes, of course, will prepare PR for this.

I mean that having to release Infima and then Docusaurus adds DX friction. If we were able to modify Infima + Docusaurus and see the live result of those changes in a single atomic PR, that would be much more convenient to review and ensure that the Infima side + Docusaurus side are both compatible with each other. 

A scenario that happens from time to time is that you change Infima, release it, then try to integrate the changes on Docusaurus, figure out Infima changes were not good enough, have to release Infima again with new changes to only then have everything working in Docusaurus.

PR #152, opened 2021-07-09
refactor: replace dist in git by diff comment

slorber:
We added dist in git so that we can more easily review the impact css processing pipeline changes (postcss etc) 

But in practice it's creating a lot of annoying conflicts in recent PRs that are not even updating the postcss setup.

This PR try to use another approach


slorber:
👍  should make the review easy without generating git conflicts @lex111 

image

lex111:
Awesome idea, thanks!

yangshun:
This is a great idea! :)

PR #153, opened 2021-07-13
refactor: remove excess styles

lex111:
A follow-up of #150

Basically, this is superfluous code that can be safely removed.

### DIFF
--- a/packages/core/styles/components/button.pcss
+++ b/packages/core/styles/components/button.pcss
@@ -98,6 +98,7 @@
   }
 
   &--block {
+    display: block;
     width: 100%;
   }
 
--- a/packages/core/styles/components/button.pcss
+++ b/packages/core/styles/components/button.pcss
@@ -98,7 +98,6 @@
   }
 
   &--block {
-    display: block;
     width: 100%;
   }
 
--- a/packages/core/styles/components/menu.pcss
+++ b/packages/core/styles/components/menu.pcss
@@ -55,7 +55,6 @@
   &__link {
     border-radius: 0.25rem;
     color: var(--ifm-menu-color);
-    cursor: pointer;
     display: flex;
     justify-content: space-between;
     line-height: 1.25;
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -36,7 +36,6 @@ html[data-theme='dark'],
   height: var(--ifm-navbar-height);
   padding: var(--ifm-navbar-padding-vertical)
     var(--ifm-navbar-padding-horizontal);
-  width: 100%;
 
   @media print {
     display: none;
@@ -148,12 +147,9 @@ html[data-theme='dark'],
 
   &__link {
     color: var(--ifm-navbar-link-color);
-    cursor: pointer;
     font-weight: var(--ifm-font-weight-semibold);
     padding: var(--ifm-navbar-item-padding-vertical)
       var(--ifm-navbar-item-padding-horizontal);
-    position: relative;
-    text-decoration: none;
 
     &:hover,
     &--active {
--- a/packages/core/styles/content/image.pcss
+++ b/packages/core/styles/content/image.pcss
@@ -10,8 +10,6 @@
 }
 
 img {
-  border-style: none;
-  box-sizing: content-box;
   max-width: 100%;
 }
 
--- a/packages/core/styles/content/table.pcss
+++ b/packages/core/styles/content/table.pcss
@@ -26,7 +26,6 @@ table {
   display: block;
   margin-bottom: var(--ifm-spacing-vertical);
   overflow: auto;
-  width: 100%;
 
   thead tr {
     border-bottom-color: var(--ifm-table-border-color);
--- a/packages/core/styles/content/typography.pcss
+++ b/packages/core/styles/content/typography.pcss
@@ -77,18 +77,5 @@ hr {
   border-color: var(--ifm-hr-border-color);
   border-style: solid;
   border-width: var(--ifm-hr-border-width);
-  box-sizing: content-box;
   margin: var(--ifm-hr-margin-vertical) 0;
-  overflow: hidden;
-
-  &:before {
-    content: '';
-    display: table;
-  }
-
-  &:after {
-    clear: both;
-    content: '';
-    display: table;
-  }
 }


packages/core/styles/components/menu.pcss
```diff
@@ -55,7 +55,6 @@
   &__link {
     border-radius: 0.25rem;
     color: var(--ifm-menu-color);
-    cursor: pointer;
```
slorber, line 58: Also assume that we apply that class to an anchor but it may not always be the case?

slorber (commented):
I'm not about removing those styles, but it could have unintended side-effects in some places.

@yangshun any opinion?

packages/core/styles/components/button.pcss
```diff
@@ -98,7 +98,6 @@
   }
 
   &--block {
-    display: block;
```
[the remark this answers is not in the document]
slorber: + the button is by default "inline-block".

This means `<button className="button button--block">` will now be inline block, that does not seem to be the purpose of this class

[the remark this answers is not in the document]
lex111: And yes and no, it seems to me that "block" in this case points more to full-width rather then `display:block` declaration.

packages/core/styles/components/menu.pcss
lex111: The point is that this class always applies to links that already have cursor pointer by default

yangshun (changes requested):
I don't think we should be removing these. I'm not confident it won't change things. Sometimes even though the styles are present in the default browser styling, we want to be defensive and still apply these styles. I'm also not convinced we did thorough testing to ensure nothing broke. A lot of these styles were written by referencing GitHub and Bootstrap's CSS and they are probably there for good reason, maybe for some legacy browser or something. I think the risk is not worth it.

packages/core/styles/content/image.pcss
```diff
@@ -10,8 +10,6 @@
 }
 
 img {
-  border-style: none;
-  box-sizing: content-box;
```
lex111, line 14: Why is it necessary for images? Is there use case for this?

packages/core/styles/content/image.pcss
```diff
@@ -10,8 +10,6 @@
 }
 
 img {
-  border-style: none;
```
lex111, line 13: This was relevant for IE9, but we don't even support IE11. So why this declaration?

packages/core/styles/components/navbar.pcss
```diff
@@ -148,12 +147,9 @@ html[data-theme='dark'],
 
   &__link {
     color: var(--ifm-navbar-link-color);
-    cursor: pointer;
     font-weight: var(--ifm-font-weight-semibold);
     padding: var(--ifm-navbar-item-padding-vertical)
       var(--ifm-navbar-item-padding-horizontal);
-    position: relative;
-    text-decoration: none;
```
lex111, line 156: We have `text-decoration:none` by default for all links, which is completely unnecessary (see https://github.com/facebookincubator/infima/blob/master/packages/core/styles/content/typography.pcss#L11).

packages/core/styles/components/navbar.pcss
```diff
@@ -148,12 +147,9 @@ html[data-theme='dark'],
 
   &__link {
     color: var(--ifm-navbar-link-color);
-    cursor: pointer;
     font-weight: var(--ifm-font-weight-semibold);
     padding: var(--ifm-navbar-item-padding-vertical)
       var(--ifm-navbar-item-padding-horizontal);
-    position: relative;
```
lex111, line 155: Leftover from already removed code when we used absolute positioning inside dropdown.

packages/core/styles/components/navbar.pcss
```diff
@@ -148,12 +147,9 @@ html[data-theme='dark'],
 
   &__link {
     color: var(--ifm-navbar-link-color);
-    cursor: pointer;
```
lex111, line 151: This is default declaration for links in all browsers, why do we need to duplicate it?

packages/core/styles/components/navbar.pcss
```diff
@@ -36,7 +36,6 @@ html[data-theme='dark'],
   height: var(--ifm-navbar-height);
   padding: var(--ifm-navbar-padding-vertical)
     var(--ifm-navbar-padding-horizontal);
-  width: 100%;
```
lex111, line 39: For `navbar__inner` we already set width to 100%, so there is absolutely no need to set full width for parent `navbar` element.

packages/core/styles/components/button.pcss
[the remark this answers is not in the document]
lex111: All right, will revert this.

packages/core/styles/content/table.pcss
```diff
@@ -26,7 +26,6 @@ table {
   display: block;
   margin-bottom: var(--ifm-spacing-vertical);
   overflow: auto;
-  width: 100%;
```
lex111, line 29: For tables we use block model, so `width:100%` makes no sense, since tables will not be full width by default. If we want tables to stretch to full width (100%) always, we need to remove `display:block`. What should we do then?

packages/core/styles/content/typography.pcss
```diff
@@ -77,18 +77,5 @@ hr {
   border-color: var(--ifm-hr-border-color);
   border-style: solid;
   border-width: var(--ifm-hr-border-width);
-  box-sizing: content-box;
   margin: var(--ifm-hr-margin-vertical) 0;
-  overflow: hidden;
```
lex111, line 82: Styles for `<hr>` scare me. Why are there so many unnecessary styles? If you remove them, nothing will change anyway. Even Bootstrap doesn't have that:

```
hr {
  margin: 1rem 0;
  color: inherit;
  background-color: currentColor;
  border: 0;
  opacity: 0.25;
}
```

lex111 (commented):
As always, I want to make sure that we use styles we _really_ need for elements. For example, why would we want to set styles that essentially duplicate the browser ones? There's actually very low risk here. That's why I've commented on each change in this PR, hopefully to convince you that removing those styles won't break anything.

slorber:
From my perspective, removing styles represent a breaking change, and risk, and has a minor ROI value for the end-users (perf micro-optimization). In general, I'd like to avoid such changes unless it's fixing an existing problem user reported.

However if styles are inspired by Bootstrap (which I didn't know) and Bootstrap decided later to remove some unused style because it's not unnecessary, I'm not against removing those styles as well, as it seems a safe thing to do (same ROI, less risk)

lex111:
I'm sure that's not the case, the current changes will not affect users websites, as I wrote above in my review, these are useless styles.

packages/core/styles/components/menu.pcss
slorber: There's no guarantee that users will apply that class to only anchors, but this is also unlikely so it should be fine

packages/core/styles/components/navbar.pcss
slorber: ok for me

packages/core/styles/components/navbar.pcss
slorber: this is the default for anchors, but users may do things that you don't expect (on their homepage or other places), like `<div className=menu__link">` and those users would be affected. It's a common practice to try to make non-link elements look like links.

Unlikely to be the case so let's remove it anyway

packages/core/styles/components/navbar.pcss
slorber: 👍

packages/core/styles/components/navbar.pcss
slorber: same as cursor: pointers, assumes class is always applied to an anchor. But seems fine to remove

packages/core/styles/content/image.pcss
slorber: 👍

packages/core/styles/content/image.pcss
slorber: 👍

packages/core/styles/content/table.pcss
slorber: Not sure to understand but this removal seems fine

packages/core/styles/content/typography.pcss
slorber: 👍

slorber (approved):
Going to approve this, feel free to merge it
Not agreeing 100% with all the changes, but they seem safe enough.

Still think this kind of PR is complicated to handle for reviewers (risky, time-consuming, little ROI), so I'd prefer to not have too much of them in the future 😅

PR #154, opened 2021-07-13
refactor: avoid using important rule in utility classes

lex111:
By nature, CSS utility classes should not use an `important` rule, because they serve as a shorthand way to add CSS declarations. Therefore, we are better removing this unnecessary rule, which has no effect to final result. At least, nothing will change in Infima/Docusaurus styles after that.

### DIFF
--- a/packages/core/styles/layout/spacing.pcss
+++ b/packages/core/styles/layout/spacing.pcss
@@ -10,23 +10,23 @@
     (0, 0.25rem, 0.5rem, 1rem, 2rem, 5rem)
   {
     .$(type)--$(size) {
-      $(type): $(value) !important;
+      $(type): $(value);
     }
 
     @each $side in (top, left, bottom, right) {
       .$(type)-$(side)--$(size) {
-        $(type)-$(side): $(value) !important;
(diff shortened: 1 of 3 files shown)


yangshun:
No I think we should keep it. When I use a utility class, I expect the
styling to be applied. If you remove the important, classes with higher
specificity might win and make the utility class useless.

FWIW, Bootstrap does this.

lex111:
On the one hand this makes sense, but for example Tailwind does not use `important` rule in their numerous utility classes. In our case it is also completely unnecessary, only complicates CSS code and increases CSS bundle in vain.

Think that `important` rule should be used wisely, and in small framework like Infima, which is basically designed to work with Docusaurus, using an `important` rule seems like completely unnecessary thing. But okay, I'll close it.

yangshun:
Glad I'm not crazy in my reasoning. Tailwind doesn't need important because
all the classes are utility classes with the same specificity. In
Infima/Bootstrap, people can combine normal CSS classes with utility
classes and there will be specificity wars. So using important is the right
way to go for our way of writing CSS.

Issue #55, opened 2020-12-02
Restyle admonitions


yangshun:
Refer to original issue - https://github.com/facebook/docusaurus/issues/2930 (I'm unable to move issues across organizations).

alexkrolick:
I'll just add that the current styling in facebook/docusaurus#2930 fails contrast ration checks quite badly, so it would be nice to come up with an alternative. See check below - Bootstrap's Alert panel (top) has a ration of 7.2, while the Docusaurus panels are in the 1.1-1.7 range. WCAG AA is 4.5 and AAA is 7. https://contrastchecker.com/?c=ffffff&b=ffba00

<img width="923" alt="Screen Shot 2020-12-21 at 5 18 41 PM" src="https://user-images.githubusercontent.com/1571667/102838049-0d38d480-43b2-11eb-8247-03ebf6a28ed2.png">

slorber:
We'll also have to ensure these issues get fixed by the redesign:
- https://github.com/facebook/docusaurus/issues/3923
- https://github.com/facebook/docusaurus/issues/2609

Steviebaa:
Regarding [docusaurus/issues/2930](https://github.com/facebook/docusaurus/issues/2930). Could there be an option similar to [Material-UI variant prop for elements](material ui button link). I.e if you chose `contained` you would get the original design or provide `outlined` or something along those lines to get the proposed style in the above issue. I kind of like both styles.

Looking forward to this one. Only thing holding me back from a release. Thanks!

alexkrolick:
Even if you want to create a variant with brighter backgrounds, the text color should have high contrast against it. In most cases the original background colors would need black text for that to be the case.

HonkingGoose:
## Introduction

> I'll just add that the current styling in facebook/docusaurus#2930 fails contrast ratio checks quite badly, so it would be nice to come up with an alternative.

The colored background also gets too much visual weight.

> Even if you want to create a variant with brighter backgrounds, the text color should have high contrast against it. In most cases the original background colors would need black text for that to be the case.

We don't need to use black/gray text to get accessible contrast. 😉 
The trick is to put dark colored text on a soft colored background.

Steve Schoger's advice

> Achieving an accessible contrast ratio is very difficult when using white text on a colored background.
> Using dark colored text on a soft colored background is much easier to make accessible, and usually looks better to boot

Source for quote: https://twitter.com/steveschoger/status/1045371124274483200

## Mock-up

[...]

PR #155, opened 2021-07-20
Alert / Admonition redesign

slorber:
Fix https://github.com/facebookincubator/infima/issues/55

## Changes

Redesign Infima alerts so that we can redesign Docusaurus admonitions:

- Reuse existing Infima colors
- Add new color variants `--ifm-color-$(color)-contrast background + `--ifm-color-$(color)-contrast-emphasis`
- Emphasis on background pass WCAG contrast tests with AAA
- Keep dark mode support
- Make it easy to customize alerts/admonitions through CSS variables
- Also redesign blockquotes to be consistent with alert design (and fix highly inaccessible contrast)
- Support for tabs

Not directly related:

- Simplify existing verbose color declaration with `@each`
- Apply missing `.clean-btn` class on Alert doc page, issue appearing after #150
- Add anchor links to Infima demo sections

## To review

- Alerts: https://deploy-preview-155--infima.netlify.app/demo/#alerts
- Alerts dark: https://deploy-preview-155--infima.netlify.app/demo/?dark=true#alerts
- Alerts in Docusaurus: https://deploy-preview-155--infima.netlify.app/docs/components/alert
- Quotes: https://deploy-preview-155--infima.netlify.app/demo/#typography
- Quotes dark: https://deploy-preview-155--infima.netlify.app/demo/?dark=true#typography

## Screenshots

Will try to keep them up-to-date with current code:

### Infima alerts

image

image

### Docusaurus admonitions

image

image

### Quotes

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# [Infima](https://facebookincubator.github.io/infima/) &middot; [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/facebookincubator/infima/blob/master/LICENSE) [![npm version](https://img.shields.io/npm/v/infima.svg?style=flat)](https://www.npmjs.com/package/infima)
+# [Infima](https://infima.dev/) &middot; [![GitHub license](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/facebookincubator/infima/blob/master/LICENSE) [![npm version](https://img.shields.io/npm/v/infima.svg?style=flat)](https://www.npmjs.com/package/infima)
 
 > ⚠️ Infima is not yet ready for production use and is being developed alongside [Docusaurus 2](https://docusaurus.io/).
 
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2364,6 +2364,9 @@ const myFun = (x, y) =&gt; {
                   when it wraps. Oh boy let's keep writing to make sure this is
                   long enough to actually wrap for everyone. Oh, you can
                   <em>put</em> <strong>Markdown</strong> into a blockquote.
+                  Here's a
+                  <a href="#">link to <code>somewhere</code> else</a> and an
+                  <code>inline code block</code>
                 </p>
               </blockquote>
             </div>
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2811,6 +2811,7 @@ let myFun = (x, y) =&gt; {
             colorElement.style.cursor = "pointer";
             colorElement.addEventListener('click', () => {
               // Not sure why, need to redo that to get the good color reliably (otherwise it may copy #ffffff)
+              // Looks related to using themed css variables: the good color can only be obtained after stylesheet has loaded?
               const clickedColor = rgba2hex(computedStyle.backgroundColor);
               copyText(clickedColor).catch(console.error)
               console.log(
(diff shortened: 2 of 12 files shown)


HonkingGoose:
This looks way better than what we have now! 🚀 

I did notice 2 things though...

## Edge case fixed with new design?

Does the new design fix this edge-case?

> Another edge case admonitions redesign should take into account:
> 
> image

## X text/button to close alert lacks contrast

Lowest contrast score for the X text on colored backgrounds is `2.49:1` in the non-active state.

> **Check the contrast for all icons.**
>
> Level AA compliance requires a contrast ratio of 3.0:1.

https://www.a11yproject.com/checklist/#color-contrast

**Steps to reproduce:**

1. Use latest Firefox browser
2. Open the preview https://deploy-preview-155--infima.netlify.app/demo/#alerts
3. Turn on developer toolbar with <kbd>F12</kbd>
4. Use the accessibility picker to hover over the X text of the alert
5. See contrast check fails AA contrast checks for the X text on all alerts

Chrome probably also has a accessibility tab with tools to detect the contrast of a element.

slorber:
Thanks @HonkingGoose , that should be better now. Also added the tabs in alerts edge case to the demo page

HonkingGoose:
> Thanks @HonkingGoose , these should be better now.

The close button is way more visible now! 😄

> Added the tabs in alerts edge case to the demo page and screenshots.

The alert component with tabs in them don't have a close button **X**, is that intended behavior?

slorber:
@HonkingGoose not intended but not important. The close button is optional and Docusaurus doesn't even use it.

HonkingGoose:
> Added a little hover effect on links, not very visible through but can't make it more visible without a lower contrast ratio, so it seems better to keep the link text readable.

That hover effect is _very_ subtle. I think it needs to be a lot clearer. 😉 

Is it an option to make the link text a bit bolder when you hover over it?
Something like this, but in Infima:

```css
a:hover {
  font-weight: 700;// bold
}
```

slorber:
> Is it an option to make the link text a bit bolder when you hover over it?

Not a good idea because link width will increase and this triggers potential layout changes on the alert (could create new lines)

I don't have much ideas how to make it more visible, because using a darker/lighter color is not very visible. Some things remain possible, like using `text-shadow`...

HonkingGoose:
> Some things remain possible, like using `text-shadow`...

I tried adding `text-shadow` to `a` tags in the browser, and it looks bad, and makes the text harder to read. 😞 

---

I did try this in the browser, which looks nice:

```css
a:hover {
  text-decoration-thickness: 2px,
}
```

But this is probably also a bad idea, because of possible layout shifts?

slorber:
Thanks, will take a look at that solution.

Using link opacity is not too great for text selection, as @Simek mentioned, so I'd rather find another way:

image

slorber:
Seems to work fine. Not perfect but a good enough solution.

Which is better?

image

image

HonkingGoose:
> Which is better, on link :hover ?
> 
> image
> 
> image

I like the bottom variant the best, as it uses a dark hue of the background color. This way it feels like the underline belongs to the component.

slorber:
> I like the bottom variant the best, as it uses a dark hue of the background color. This way it feels like the underline belongs to the component.

Let's test this then, not too hard to revert if people disagree

---

@lex111 @yangshun I don't like too much having this inline code block horizontal padding, as it breaks the link underline and separates too much the word from the rest.

image

Going to reduce it, let me know if that's a problem:

image

PR #156, opened 2021-07-21
chore: use correct website URL in Readme

Simek:
Currently used link leads to the broken page:

Screenshot 2021-07-21 123313


slorber:
thanks :)

PR #157, opened 2021-07-25
refactor: new blockquotes styling + minor tweaks

lex111:
Currently, blockquotes are very similar to secondary alerts, so it would be good to make them more different in styling.
Similar to v1, we can take light yellow color as background, then current content color (black/white depending on color mode) and also remove border-radius (unlike in alerts). We'll end up with something like this:

| Before   | After    |
| -------- | -------- |
| image | image |

Beside this, as part of this PR, emphasis word is renamed to "foreground"  (for better clarity), and removing increase of text decoration line at focus state, because it does not make much sense.

### DIFF
--- a/packages/core/styles/components/alert.pcss
+++ b/packages/core/styles/components/alert.pcss
@@ -51,6 +51,26 @@
   color: var(--ifm-alert-foreground-color);
   padding: var(--ifm-alert-padding-vertical) var(--ifm-alert-padding-horizontal);
 
+  &__heading {
+    align-items: center;
+    display: flex;
+    font: bold var(--ifm-h5-font-size) / var(--ifm-heading-line-height)
+      var(--ifm-heading-font-family);
+    margin-bottom: 0.5rem;
+    text-transform: uppercase;
+  }
+
+  &__icon {
+    display: inline-flex;
+    margin-right: 0.4em;
+
+    svg {
+      fill: var(--ifm-alert-foreground-color);
+      stroke: var(--ifm-alert-foreground-color);
+      stroke-width: 0;
+    }
+  }
+
   .close {
     color: var(--ifm-alert-foreground-color);
     margin: calc(var(--ifm-alert-padding-vertical) * -1)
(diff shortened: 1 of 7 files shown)


lex111:
I also added markup and styles for alert heading. However, I'm not sure we need to add styles for SVG icon or/and markup for it at all to Infima codebase. Perhaps it's better to do it in the remark-admonition plugin styles?

image

packages/core/styles/content/typography.pcss
```diff
@@ -94,8 +94,8 @@ blockquote {
 
   a {
     text-decoration-color: var(--ifm-blockquote-border-color);
-    &:hover,
-    &:focus {
+
+    &:hover {
```
slorber, line 98: oh I see what you mean now, I thought you wanted to remove the :hover too

packages/core/styles/components/alert.pcss
```diff
@@ -46,11 +48,31 @@
   border-left-width: var(--ifm-alert-border-left-width);
   border-radius: var(--ifm-alert-border-radius);
   box-shadow: var(--ifm-alert-shadow);
-  color: var(--ifm-alert-emphasis-color);
+  color: var(--ifm-alert-foreground-color);
   padding: var(--ifm-alert-padding-vertical) var(--ifm-alert-padding-horizontal);
 
+  &__heading {
```
slorber, line 54: > I also added markup and styles for alert heading. However, I'm not sure we need to add styles for SVG icon or/and markup for it at all to Infima codebase. Perhaps it's better to do it in the "remark-admonitions" plugin styles as like before?

That seems fine to have it there, we can see later if we find a way to use those classes in the remark-admonitions plugin, or eventually remove this CSS if useless

slorber (approved):
👍 looks like changes we already agreed on

dsmmcken:
Didn't this just trade it looking like secondary-alerts to warning-alerts? I prefer this, but any reason to not just use the existing -warning colors rather than introduce three new yellow hardcodes? Related to #163 
image

slorber:
The goal is to make quotes not look like alerts. We probably don't want to use the exact same color, on purpose

PR #161, opened 2021-07-29
refactor: minor navbar/dropdown cleanup

lex111:
First of all Docusaurus currently uses not entirely correct markup for dropdown navbar item, because `navbar__item` class is used for link inside:

image

However on demo page shows that `navbar__item` class should be on parent element only:

image

It follows that `navbar__link` class should not add padding, since `navbar__item` class is designed to do this. That's why we now have this small duplication of code:

image

Also, we don't have to remove padding from dropdown navbar item because we can simply adjust `top` position.

### DIFF
--- a/packages/core/styles/components/dropdown.pcss
+++ b/packages/core/styles/components/dropdown.pcss
@@ -48,8 +48,8 @@
     overflow-y: auto;
     padding: 0.5rem;
     position: absolute;
-    top: 100%;
-    transform: translateY(-10px);
+    top: calc(100% - var(--ifm-navbar-item-padding-vertical));
+    transform: translateY(-0.625rem);
     visibility: hidden;
     z-index: var(--ifm-z-index-dropdown);
     @mixin transition opacity transform visibility;
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -133,8 +133,6 @@ html[data-theme='dark'],
       var(--ifm-navbar-item-padding-horizontal);
 
     &.dropdown {
-      padding: 0;
-
       ^^&__link:not([href]) {
         pointer-events: none;
       }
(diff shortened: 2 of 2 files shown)


packages/core/styles/components/dropdown.pcss
```diff
@@ -48,8 +48,8 @@
     overflow-y: auto;
     padding: 0.5rem;
     position: absolute;
-    top: 100%;
-    transform: translateY(-10px);
+    top: calc(100% - var(--ifm-navbar-item-padding-vertical));
+    transform: translateY(-0.625rem);
```
lex111, line 52: Just for better consistency, since we are using `rem` rather than `px`.

slorber: I don't understand where this value 0.625 comes from?

lex111: Just converted 10px to 0.625rem (10 / 16).

slorber:
👍

slorber:
> It's also pretty weird that they use full width mobile sidebar, because we don't have explicit button to close it, which means that user can't close mobile sidebar unless they clicks somewhere.


Yes, also mentioned that.

I'm wondering if we shouldn't add a close button btw. Clicking on the empty overlay is a common pattern but it may not be explicit enough (may even be an accessibility issue?), and using a fullscreen drawer make sense to me if you want to.

lex111:
Maybe yes, for better a11y/UX. Something like that?

image

slorber:
I find the double arrows a bit confusing.

Maybe we could add the new close button (same as the announcement bar) on the other side? (probably worth creating an `@theme/IconClose` component then)

lex111:
I have added "explicit" close button in PR #172 -- the close button is in same position as the menu icon (hamburger). However, stylistically they are slightly different icons, line thickness is different, so we need to find SVG close icon, line width which would be similar to menu button.  

image

PR #162, opened 2021-07-29
refactor: make optional sidebar back button

lex111:
Because in Docusaurus we need to display secondary menu even when there is no main one, we have to remove back button to main menu. This PR paves the way to this, setting proper padding when there is no back button:

| Before   | After    |
| -------- | -------- |
| image | image |


slorber:
👍

PR #167, opened 2021-08-26
refactor: use shorthand property to reset list

lex111:
Purely for simplicity and micro-optimization


lex111:
@slorber it's probably absolutely safe, can merge it?

PR #170, opened 2021-08-27
fix: use standard font-family in cleaned buttons

### LINKED ISSUE #169: Back to main menu => missing font-family
The CSS var does not modify its font:

image

lex111:
Fixes #169


slorber:
👍

PR #172, opened 2021-08-31
feat: add close mobile navbar button

lex111:
Based on https://github.com/facebookincubator/infima/pull/161#issuecomment-908213009

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -160,7 +160,10 @@
       <div role="presentation" class="navbar-sidebar__backdrop"></div>
       <div class="navbar-sidebar">
         <div class="navbar-sidebar__brand">
-          <a class="navbar__brand navbar-sidebar__close" href="#" role="button">
+          <button type="button" class="clean-btn navbar-sidebar__close">            
+            <svg height="20" viewBox="0 0 413.348 413.348" width="20"><path d="m413.348 24.354-24.354-24.354-182.32 182.32-182.32-182.32-24.354 24.354 182.32 182.32-182.32 182.32 24.354 24.354 182.32-182.32 182.32 182.32 24.354-24.354-182.32-182.32z"/></svg>
+          </button>
+          <a class="navbar__brand" href="#" role="button">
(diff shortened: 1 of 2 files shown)


slorber:
LGTM thanks

Let me know if you plan to add it to Docusaurus asap, or I'll do it (need for RN website)

slorber:
yes ;)

lex111:
I'm wondering if moving this button on right side, where we have decent free space, would that be better or not?

| Before   | After    |
| -------- | -------- |
| image | image |


I thought this positioning would be better for those who are used to current mobile navbar -- they will not have to get used to  new design.

However you mentioned that you would like to be able to hide close button of mobile navbar (by analogy in announcement bar, I guess), it seems like not very useful option to me. Since it is better to have two  possibilities to close mobile navbar (actually, yes, it's common thing, and therefore it was mistake not to have such button). So maybe we just need to move the button to another suitable place?

For example, [Google help docs](https://support.google.com/docs/answer/7068618?hl=en&ref_topic=9045928) uses this pattern:

image

slorber:
> I'm wondering if moving this button on right side, where we have decent free space, would that be better or not?

This is what I initially suggested, but both sides are fine to me.

> However you mentioned that you would like to be able to hide close button of mobile navbar (by analogy in announcement bar, I guess), it seems like not very useful option to me.

I think it can be useful if the site name is long and the drawer has small width: this avoids truncating the site name.

And drawers without close buttons are also common, as long as there's a visible overlay to close it. Gmail does it.

I'd provide an option like `navbar.mobileSidebarCloseButton: 'none' | 'right' | 'left'` or something, but this can be done later if someone requests it

lex111:
So I can move the close button to the right side now, if you don't mind?

We can also increase mobile navbar width slightly (`ifm-navbar-sidebar-width`)  from 80 to 85 vw to get even more space.

| Before   | After    |
| -------- | -------- |
| image  | image|

slorber:
Yes I'm fine with that

PR #173, opened 2021-09-01
chore: change CI branch from master to main

yangshun:
## Summary

The OSS team has changed all `master` branches to `main` for most FB repos and we need to update and references to `master` in our code.

Do we need to change anything on Netlify side?

## Test Plan

CI should pass.


slorber:
Did the change on Netlify, all should be fine

PR #174, opened 2021-09-01
refactor: align mobile sidebar close button to right side

lex111:
image

### DIFF
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -20,7 +20,7 @@
   --ifm-navbar-search-input-color: var(--ifm-color-emphasis-800);
   --ifm-navbar-search-input-placeholder-color: var(--ifm-color-emphasis-500);
   --ifm-navbar-search-input-icon: url('data:image/svg+xml;utf8,<svg fill="currentColor" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" height="16px" width="16px"><path d="M6.02945,10.20327a4.17382,4.17382,0,1,1,4.17382-4.17382A4.15609,4.15609,0,0,1,6.02945,10.20327Zm9.69195,4.2199L10.8989,9.59979A5.88021,5.88021,0,0,0,12.058,6.02856,6.00467,6.00467,0,1,0,9.59979,10.8989l4.82338,4.82338a.89729.89729,0,0,0,1.29912,0,.89749.89749,0,0,0-.00087-1.29909Z" /></svg>');
-  --ifm-navbar-sidebar-width: 80vw;
+  --ifm-navbar-sidebar-width: 83vw;
 }
 
 html[data-theme='dark'],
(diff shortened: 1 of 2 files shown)


packages/core/styles/components/navbar.pcss
```diff
@@ -284,7 +284,8 @@ html[data-theme='dark'],
     }
 
     &__close {
-      margin: 0 0.8rem 0 0.3rem;
+      display: flex;
+      margin-left: auto;
```
slorber, line 288: As brand and close are already in a flex row, I'm not too fan of using `margin-left: auto`.

Also why use `display: flex` here?

As we eventually plan to allow configuring the icon position later, it may be a good idea to account for that today in Infima, as it could later lead to an Infima breaking change? 

What about this instead:

```css
  &__brand {
    fiex: 1;
    margin-right: 1rem; // can now be removed? brand with no icon should take all navbar space
  }

  &__right {
    flex: 0;
    margin-left: 1rem;
  }
  &__left {
    flex: 0;
    margin-right: 1rem;
  }
```

It would be more future proof to add icons on both sides (including multiple icons)

---

In my opinion, we could even add this to Infima after it's already been in use in Docusaurus for a while, battle-tested and it's considered stable enough. This way we ensure there are limited breaking changes in Infima. It doesn't mean there won't be breaking changes in Docusaurus, but Docusaurus classes are not officially documented and harder to target with custom CSS)

slorber (changes requested):

slorber:
> `margin-left: auto` is just quick way to move flex item to the right
side. I could have used `flex: 1; justify-content: flex-end` instead
but does it really matter?

I don't think it's a good practice to mix CSS layout systems when you can stick to one.
The end result may be similar but the implementation would look cleaner (similarly, I wouldn't use `float: right`).

> `display: flex` is needed to vertical center "x" icon correctly
(missed it in the first PR).

The flex row already align row items to the center.
The alignment problem is due to the button not being a perfect square (it's `20x23` currently)

To fix it properly, you either have to:
- Define explicit `width` + `height` in the button and let the svg use 100% of it.
- Use `line-height: 0;` so that button is not larger than the svg, leading to an unexpected expanded clickable area surface

> I wouldn't separate current layout on right and left columns now,

Agree, let's do that later

---

Actually, I noticed other issues that are a bit annoying and only happen at certain viewport widths:

image

As I'd like to fix this asap for the RN website, I'll take over this PR and finish it

slorber:
Ok, so I'm going to merge this as-is because unfortunately to do what I want would require a deeper refactoring. `flex: 1` can't be applied to the site title because the class is applied directly to the link (and not a layout container) so it would create a link larger than the text with an invisible click area. Also it would move the theme switch (that we only have on Docusaurus site).

lex111:
Alright, thanks.
Agreed, exactly, resetting of line-height would be better solution. Let's to do this during another refactoring.

PR #176, opened 2021-09-15
Add support for thin scrollbars on firefox

dsmmcken:
Before in firefox:
image

After chrome/firefox:
image

### DIFF
--- a/packages/core/styles/utilities/custom-scrollbar.pcss
+++ b/packages/core/styles/utilities/custom-scrollbar.pcss
@@ -13,6 +13,9 @@
 }
 
 @media (pointer: fine) {
+  .thin-scrollbar {
+    scrollbar-width: thin;
+  }
   .thin-scrollbar::-webkit-scrollbar {
     height: var(--ifm-scrollbar-size);
     width: var(--ifm-scrollbar-size);


dsmmcken:
@lex111 just a friendly bump. Don't know if this made anyone's review queue, as no reviewer got assigned by default.

lex111:
Is it just me or does scrollbar look too thin on Linux? Let's get an opinion from @slorber.

image

HonkingGoose:
@lex111 Your scrollbar looks really thin compared to the preview image posted by the PR author.

---

Do we really _need_ a thin scrollbar, I think maybe we're better off with the full-fat scrollbar, as that is easier to grab with the mouse.

MDN Web Docs has this to say:

> Use this property with caution — setting `scrollbar-width` to `thin` or `none` can make content hard or impossible to scroll if the author does not provide an alternative scrolling mechanism. While swiping gestures or mouse wheels can enable scrolling on such content, some devices have no scroll alternative.
> 
> WCAG criterion 2.1.1 (Keyboard) has been in place for a long time to advise on basic keyboard accessibility, and this should include scrolling of content areas. And introduced in WCAG 2.1, criterion 2.5.5 (Target Size) advises that touch targets should be at least 44px in width and height (although the problem is compounded on high-resolution screens; thorough testing is advised).

Source of quote: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width

dsmmcken:
My screenshot was from Firefox + win10 for reference. Linux + Firefox does look thinner 🤷 

> Do we really need a thin scrollbar, I think maybe we're better off with the full-fat scrollbar, as that is easier to grab with the mouse.

No, but it usually looks better in context of heavily styled pages. There are existing styles for webkit that are applying a thin scrollbar, I was looking for consistency across the board.

> "While swiping gestures or mouse wheels can enable scrolling on such content, some devices have no scroll alternative."

Personal opinion, and not an actual stat, but is this still true in 2021? Speaking for my users, which are exclusively devs, I would be very surprised if any didn't.

Either way, I've already overridden the style for my site, seeing the one on the left is what made me investigate.

image

slorber:
That makes sense to me to use a thin scrollbar wherever possible for consistency, this is what `.thin-scrollbar` aims to do 🤪 

If this is annoying for some users (accessibility or Linux+Firefox), it's possible to revert this change with custom CSS.

PR #180, opened 2021-10-02
fix: use better color for active link in primary styled navbar

lex111:
Right now the navbar style looks pretty weird. I don't know what it should be, but it probably makes sense to make it closer to the dark style.

Before:

image

image

After:

image

image


slorber:
Agree, but hope it won't be annoying to anyone.

This assumes the primary color is darkish, which is not something that will always be true even if it's generally the case.

IMHO I'm not sure using a primary preset is a good thing, we should consider deprecating it because a CSS var can't easily know if a color is dark or light and infer a proper default text color: we'd rather ask Docusaurus users to explicitly provide navbar text/background colors in this case (and in JS inferring good defaults might be easier).

PR #186, opened 2021-10-17
Add project structure

fanny:
This PR adds info about the project's package which would guide new contributors to comprehend the Infima's structure.


slorber:
Thanks 👍

PR #190, opened 2021-11-26
feat: embed icon of collapsible item as separated element

### LINKED ISSUE #188: Sidebar category icon not clickable due to using ::after
The sidebar category icon is using an `::after` pseudo-element.

image

It worked fine until now, but for https://github.com/facebook/docusaurus/pull/5830 I'd like to be able to add a click listener to this icon (otherwise this category couldn't be collapsed: https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/category/guides/), and eventually add some extra `:hover` styling to make it clear that the icon can be clicked independently from the text.

We could also allow Docusaurus users to provide their own custom icon if they want to.

lex111:
Should resolve #188 

Not the ideal option (and probably not final design), but so far I can't imagine how it should look like. Definitely need to support two versions -- with pseudo element in case of categories without index page (as it was before), and with new use case, when text and icon are represented as separate elements.

Technically, for menu item with category index need to add a new `<button>` element after the link element. In the case of a "regular" category, everything is unchanged (only the `menu__link--sublist` class on menu link is required).

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -241,9 +241,10 @@
                 </a>
               </li>
               <li class="menu__list-item">
-                <a class="menu__link menu__link--sublist" href="#url">
+                <a class="menu__link" href="#url">
                   Category
                 </a>
+                <button type="button" class="clean-btn menu__caret"></button>
                 <ul class="menu__list">
                   <li class="menu__list-item">
                     <a class="menu__link menu__link--active" href="#url">
@@ -744,9 +745,10 @@
                     </a>
                   </li>
                   <li class="menu__list-item">
-                    <a class="menu__link menu__link--sublist" href="#url">
+                    <a class="menu__link" href="#url">
                       Category
                     </a>
+                    <button type="button" class="clean-btn menu__caret"></button>
                     <ul class="menu__list">
                       <li class="menu__list-item">
                         <a class="menu__link menu__link--active" href="#url">
@@ -766,9 +768,10 @@
                     </ul>
                   </li>
                   <li class="menu__list-item menu__list-item--collapsed">
-                    <a class="menu__link menu__link--sublist" href="#url">
+                    <a class="menu__link" href="#url">
                       A Very Very Very Very Very Long Category
                     </a>
+                    <button type="button" class="clean-btn menu__caret"></button>
                     <ul class="menu__list">
                       <li class="menu__list-item">
                         <a class="menu__link" href="#url">Second Level</a>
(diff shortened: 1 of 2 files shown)


Josh-Cena:
> Definitely need to support two versions -- with pseudo element in case of categories without index page (as it was before), and with new use case, when text and icon are represented as separate elements.

IMO we should make the category dropdown separate elements in all cases to keep consistency. This could (a) allow users to swap out the chevron with other icons and (b) make redesigning the UX more convenient in the future if there's ever such a need. (more control on the Docusaurus side rather than having to manipulate a monolithic element)

lex111:
There is no need to always have two separate elements if they are doing the same thing. It's confusing and doesn't make sense.

> nuance in the background color when hover

I don't really understand what exactly is wrong?

I would still embed the icon through CSS -- users can replace it with their own in a custom CSS file by targeting on pseudo element, or override default value of CSS var `--ifm-menu-link-sublist-icon`.

Josh-Cena:
> I don't really understand what exactly is wrong?

No, before I thought all category links have the same UI, just those with an index have a different clickable chevron, so I suggested using a unified CSS. Now I realized the two have a whole different styling, it's fine to keep it that way.

> users can replace it with their own in a custom CSS file by targeting on pseudo element

It's slightly different from the usual practice on the Docusaurus site, but yes we can educate the users. Maybe it's an even better approach because of reusability? (cf https://github.com/facebook/docusaurus/issues/5865)

slorber:
> I would still embed the icon through CSS -- users can replace it with their own in a custom CSS file by targeting on pseudo element, or override default value of CSS var --ifm-menu-link-sublist-icon.

I'm not sold on what exactly we should do about icons yet, but in any case we'd rather enable a consistent way to override icons, not half CSS half swizzle. 

But for now this approach can be fine and unlocks me. We'll think about icons in general as part of https://github.com/facebook/docusaurus/issues/5865

Josh-Cena:
In terms of the design with separate elements, what about the GitBook design? https://docs.gitbook.com/organizations/billing-settings

The category title still as a whole acts as a link, but the chevron is a little clickable button embedded

slorber:
I don't have a strong opinion on that, we can polish the design later as long as the UX is complete and the design is good enough

This active category design surprised me a bit (not in the current demo) but I don't dislike it:

image

slorber:
integration can be seen at https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/category/guides/

lex111:
> This active category design surprised me a bit (not in the current demo) but I don't dislike it:

You merged too soon... well, never mind. here's revised design, improved to work with hover/active state.  Is that better?

ezgif com-gif-maker (11)

slorber:
Thanks :) yes that seems better to me.

> You merged too soon... well, never mind. here's revised design, improved to work with hover/active state. Is that better?

If you still want to work on the PR, it's better for me if you mark it as draft, otherwise I'll consider ready to merge from your side :)

PR #196, opened 2021-12-13
feat: make pagination-nav arrows stylizable

noomorph:
**This issue stems from:** https://github.com/facebook/docusaurus/pull/6053

Short summary:

I'd like to stylize the `pagination-nav` so it looks like this:

image

The problem is that the arrows `»` are just inlined into the markup and cannot be hidden via CSS:

image

This is why I move them to `::before` and `::after` pseudo-elements, so that later on I am able to hide them completely and continue my styling described above.

### DIFF
--- a/packages/core/styles/components/pagination-nav.pcss
+++ b/packages/core/styles/components/pagination-nav.pcss
@@ -21,11 +21,11 @@
     display: flex;
     flex: 1 50%;
     max-width: 50%;
-    
+
     &--next {
       text-align: right;
     }
-    
+
     & + & {
       margin-left: var(--ifm-spacing-horizontal);
     }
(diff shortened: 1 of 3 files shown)


packages/core/styles/components/pagination-nav.pcss
```diff
@@ -38,19 +38,27 @@
...
     font-weight: var(--ifm-heading-font-weight);
     word-break: break-word;
   }
-  
+
+  &__item:first-child &__label::before {
+    content: '« '
+  }
+
+  &__item--next &__label::after {
```
lex111, line 58: Same as above -- put it inside into the already existing rule on 25 line.

packages/core/styles/components/pagination-nav.pcss
```diff
@@ -38,19 +38,27 @@
...
-  
+
   &__label {
     font-size: var(--ifm-h4-font-size);
     font-weight: var(--ifm-heading-font-weight);
     word-break: break-word;
   }
-  
+
+  &__item:first-child &__label::before {
```
lex111, line 54: Add this inside existent rule on 20 line. You can use `^&` selector to reference on  `pagination-nav` class (see https://github.com/toomuchdesign/postcss-nested-ancestors for more details).

noomorph:
@lex111 I followed the suggestion with `^` selector, thanks, but maybe I'll disagree with the suggested relocation itself. Could you check out my PR once again? Doesn't it make sense to put `::before` and `::after` under the scope of the parent element?

lex111 (approved):
It's all right, thanks!

slorber:
Hey

I'm not against this change, but IMHO it won't solve your use-case

At most you would be able to do:

image

And not:

image

(which should rather be handled with swizzle IMHO)

@lex111 feel free to merge if you consider it's a good improvement

noomorph:
@slorber well, I'm just disabling those `::after, ::before` and adding them (with a background-image) to a parent container.

lex111:
Yep, it's still a good improvement since it allows to control the arrow in pagination links.

PR #204, opened 2022-02-11
refactor: use semi-transparent background for striped rows

### LINKED ISSUE #195: Table stripe color should be semi-transparent
This doesn't look right:

image

I'm also trying to allow users to tweak the site's background color: https://github.com/facebook/docusaurus/pull/5822 and any other color doesn't look right with the table stripes still staying dull-black.

lex111:
Fixes #195

@Josh-Cena take a look please.

### DIFF
--- a/packages/core/styles/common/dark-mode.pcss
+++ b/packages/core/styles/common/dark-mode.pcss
@@ -38,7 +38,7 @@ html[data-theme='dark'] {
   --ifm-scrollbar-thumb-background-color: #686868;
   --ifm-scrollbar-thumb-hover-background-color: #7a7a7a;
 
-  --ifm-table-stripe-background: rgba(255, 255, 255, 0.05);
+  --ifm-table-stripe-background: rgba(255, 255, 255, 0.07);
 
   @each $color in (primary, secondary, success, info, warning, danger) {
     --ifm-color-$(color)-contrast-background: color-mod(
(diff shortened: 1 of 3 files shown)


Josh-Cena (approved):
That looks nice! I wonder if it can be slightly _less_ transparent so as to retain contrast in dark mode, but it looks much better

lex111:
Like this?

| Before   | After    |
| -------- | -------- |
| image |  image |

Josh-Cena:
Yes, that looks good 👍

slorber:
sorry missed this PR since I don't use email notifications anymore 😅 

That looks fine 👍

packages/core/styles/common/dark-mode.pcss
```diff
@@ -38,6 +38,8 @@ html[data-theme='dark'] {
   --ifm-scrollbar-thumb-background-color: #686868;
   --ifm-scrollbar-thumb-hover-background-color: #7a7a7a;
 
+  --ifm-table-stripe-background: rgba(255, 255, 255, 0.07);
```
slorber, line 41: Was wondering if we ever used this `default-dark/default-dark.pcss` file @lex111 @yangshun ?

https://deploy-preview-204--infima.netlify.app/demo/css/default-dark/default-dark.css

As far as I understand it's not used by Docusaurus and it's not either really maintained (for example current PR does not update it, and probably many others too)

Maybe we could simply delete this entrypoint?

slorber: Docusaurus is using this file only:

```js
function getInfimaCSSFile(direction: string) {
  return `infima/dist/css/default/default${
    direction === 'rtl' ? '-rtl' : ''
  }.css`;
}
```

PR #206, opened 2022-02-15
fix: prefer background-color property (#198)

### LINKED ISSUE #198: Use of `background` property to set background color clobbers other background properties
I've run into some trouble customizing a Docusaurus site due to Infima's pattern of specifying the entire `background` property for elements when it's actually just targeting the `background-color`.

The specific case I'm bumping against is [here](https://github.com/facebookincubator/infima/blob/8c592e4fad80d410fcb9ffb132e6b2efc70c9db4/packages/core/styles/components/menu.pcss#L132), but I did a quick search and see many other instances where the `background` property is being set to just a color.

**Problem:** This means any downstream customizing such as setting a background image gets clobbered unnecessarily.

**Solution:** Unless Infima specifically _means_ to clobber any other background properties, it should just set the `background-color` when that's all that's being specified.

I'm happy to submit a PR for this, although I'm unsure if this is a change I should make everywhere I find it or just in that one instance. Guidance welcome!

reefdog:
Fixes #198 

As described in the linked issue, using the `background` CSS property to set only the `background-color` makes it more difficult to set other individual `background-*` properties downstream, since they get clobbered unless the selector chain is specific enough.

As a rule, it would be better to use the `background-color` property specifically when that is all that is being set. Per @lex111's advice, this MR only updates the example I am having problems with.

Note that this does not completely resolve the linked issue, but it does resolve _my_ particular use case.

Since this is my first PR to the project, please let me know if there are additional changes I should make!

### DIFF
--- a/packages/core/styles/components/menu.pcss
+++ b/packages/core/styles/components/menu.pcss
@@ -129,7 +129,7 @@
       }
 
       &:not(^&--sublist) {
-        background: var(--ifm-menu-color-background-active);
+        background-color: var(--ifm-menu-color-background-active);
       }
     }
   }


lex111 (approved):

lex111:
@reefdog great, thanks for the fix!

FYI you don't have to put the ID of issue in the PR/commit title, you can link PR with corresponding issue using special keywords (see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword for more details).

reefdog:
@lex111 Super, thanks! That issue number carried over from the issue commit title and I forgot to remove it. Will do so in future.

PR #212, opened 2022-03-11
Revert "Update dependencies (#211)"

slorber:
This reverts commit 785ce1ee077e52ea6c0514297323666d66050fb5.

See regression here: https://github.com/facebookincubator/infima/pull/211#discussion_r823828062

@RDIL feel free to re-open a PR if you know how to upgrade the deps without changing the existing behavior

I'm reverting everything here for clarity/simplicity but there's probably a few things that we can upgrade safely (node version, prettier etc)


slorber:
looks like the CSS dif bot finally works, it probably only fails on a fork, particularly when @RDIL submitted the PR from the main branch so it probably attempted to compare "fork main vs fork main" somehow 😅

Josh-Cena:
@slorber See https://github.com/csstools/postcss-plugins/issues/277

Since `color-mod` is not there anymore, do you think we should just hardcode our colors instead?

yangshun:
Yes I think so. I think using `color-mod`/any formulaic approach is not the best way. 

We could pick some colors from [Tailwind](https://tailwindcss.com/docs/customizing-colors#default-color-palette) or [Radix UI](https://www.radix-ui.com/colors)

slorber:
> as the adjustments need not be the same for each color... 

Afaik there are a few interesting things happening regarding CSS colors recently, that's worth looking at the new shiny things and see if one or a combination could solve our use-case in a generic and maintainable way 🤔

PR #213, opened 2022-03-12
refactor: more emphasized background color for table header

pranabdas:
I think table header looks better with more emphasized background (especially because the first tbody row is less emphasized). Intended result: 

Before | After
------ | -----
before-light | after-light
before-dark | after-dark


slorber:
👍 agree it seems better

we could even have a different background color for the header maybe?

going to merge this, but open for further improvements

PR #214, opened 2022-03-18
refactor: better positioning of menu item icon

lex111:
The positioning of a caret down icon of menu item  to the right should not affect other elements, in particular a external icon.


slorber:
👍 seems a better default

PR #215, opened 2022-03-19
website: upgrade docusaurus and other dependencies

pranabdas:
Requires node 14 at least.


Josh-Cena:
Unfortunately, our website's usage of Prettier doesn't seem to be compatible with any newer version of it. I would refactor that in a later commit.

Josh-Cena:
Going to merge it like this. There's something not well with our diff setup, since this PR is made from a fork's `main` branch as well, similar to #211

PR #217, opened 2022-03-19
refactor: make TOC link clickable over full width

lex111:
Just expanding the clickable area of a link is good for UX.

### DIFF
--- a/packages/core/styles/components/table-of-contents.pcss
+++ b/packages/core/styles/components/table-of-contents.pcss
@@ -37,6 +37,7 @@
 
   &__link {
     color: var(--ifm-toc-link-color);
+    display: block;
 
     &:hover,
     &:hover code,


Josh-Cena:
TBH, I don't really know which one is better UX. Because our TOC doesn't have a background color signaling where the boundaries of each link lie, the current UX is also a bit weird to me.

lex111:
To click on a link in this TOC, you have to hover your cursor exactly over the text content of a link, it is not very convenient, especially in the case of TOC links with short text. 

image

This is our flaw, because in fact, this UX pattern is actively used in many large docs sites:

- https://developers.google.com/cast/docs/developers
- https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Basics
- https://www.php.net/manual/en/language.attributes.overview.php
- https://vuejs.org/guide/introduction.html

Josh-Cena:
I see. I was looking at the MDN docs and they have the outline + background color, that's why I said the lack of boundaries makes the UX a little weird... But seeing that all the other sites are taking the same approach, I think this is good 👍

PR #219, opened 2022-03-21
chore: refactor website setup

Josh-Cena:
Replaced all `<Playground>` with a remark plugin, so that we never ship Prettier to client side.

Also enabled autogenerated sidebars.

Also removed a lot of useless front matter.

Because there's absolutely no boilerplate now, the TEMPLATE has been removed.


yangshun:
> so that we never ship Prettier to client side.

Wow you're a hero!

Josh-Cena:
Thanks 😄 I had to do it anyways, since Prettier v1.17+ seems to bundle Prettier in a different way that causes the `import prettier from 'prettier/standalone'` to fail

slorber:
nice 👍 didn't even know Prettier was bundled in that playground 😅

PR #220, opened 2022-03-21
chore: upgrade dependencies

Josh-Cena:
Some trivial ones. Harder ones from #211 are not touched (yet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,10 @@
     "husky": "^7.0.4",
     "lint-staged": "^12.3.7",
     "prettier": "^2.6.0",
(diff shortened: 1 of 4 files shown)


yangshun (approved):

Josh-Cena:
Hold on, dist CSS is different

Josh-Cena:
That's better... @slorber #211 could be caught but wasn't because of a little bug in the CSS diff comment, fixed in #221

Josh-Cena:
Diff looks harmless now. Going to merge this and take care of the breaking changes in `postcss-each` and `postcss-preset-env` later

PR #221, opened 2022-03-21
chore: fix CSS diff action when upgrading deps

Josh-Cena:
Hope this would do something... Installing after building seems like a terrible idea to me


slorber:
🙈 good catch

PR #222, opened 2022-03-21
chore: upgrade postcss-preset-env to v7

Josh-Cena:
This makes things work at bare minimum.

cc @rdil @slorber We may need to eventually remove `color-mod` and hardcode color values until the spec is updated.

Related #211 #212

### DIFF
--- a/packages/postcss-preset-infima/index.js
+++ b/packages/postcss-preset-infima/index.js
@@ -10,6 +10,7 @@
 const postcssEach = require('postcss-each');
 const postcssImport = require('postcss-import');
 const postcssFor = require('postcss-for');
+const postcssColorModFunction = require('postcss-color-mod-function');
 const postcssPresetEnv = require('postcss-preset-env');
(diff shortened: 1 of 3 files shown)


slorber:
🐵 that's exactly what I tried locally on RDIL PR but it didn't work 😅 

We can keep color-mod for now, even if it's not in the spec that wouldn't be the first time someone use something not in the spec. As long as it keeps working it's probably not worth refactoring everything in a less maintainable way.

As far as I understand the CSS spec still allows to do the kind of lighten/darken modifications we want to do with "relative color syntax" and the "from" keyword so if we refactor I'd rather use this.

https://www.stefanjudis.com/notes/new-in-css-relative-colors/

slorber:
Not against but remember that this is a build step and works fine so the refactoring has a low ROI and may break things

Now if you can refactor and it produces exactly the same CSS as before in a more maintainable way, and is thus easy to review, why not 👍

PR #223, opened 2022-03-21
refactor: minor refactor for table header background

pranabdas:
Changes made: 
- <s>Distinct background color for table header</s> (see https://github.com/facebookincubator/infima/pull/213#issuecomment-1072230336)
- <s>Rows are highlighted on hover. I think helpful to follow along wide table rows with many columns.</s>
- <s>Reduce vertical padding for table rows. I think the table rows are too relaxed vertically.</s>

As suggested, the earlier changes that I proposed should be done at the user land. I reverted the changes, only a minor refactor there for the table header color.

### DIFF
--- a/packages/core/styles/content/table.pcss
+++ b/packages/core/styles/content/table.pcss
@@ -14,7 +14,7 @@
   --ifm-table-border-width: 1px;
   --ifm-table-border-color: var(--ifm-color-emphasis-300);
 
-  --ifm-table-head-background: inherit;
+  --ifm-table-head-background: rgba(0, 0, 0, 0.06);
   --ifm-table-head-color: inherit;
   --ifm-table-head-font-weight: var(--ifm-font-weight-bold);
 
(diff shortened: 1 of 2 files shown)


packages/core/styles/content/table.pcss
```diff
@@ -33,16 +33,16 @@ table {
...
-    background-color: var(--ifm-table-stripe-background);
-  }
-
   tr {
     background-color: var(--ifm-table-background);
     border-top: var(--ifm-table-border-width) solid
       var(--ifm-table-border-color);
   }
 
+  tr:hover td {
```
Josh-Cena, line 42: I don't really like this UX, particularly that the color change isn't subtle at all. This should better be done in userland, because IMO tables should usually be static.

slorber: agree, not fan

packages/core/styles/content/table.pcss
```diff
@@ -14,7 +14,7 @@
   --ifm-table-border-width: 1px;
   --ifm-table-border-color: var(--ifm-color-emphasis-300);
 
-  --ifm-table-head-background: inherit;
+  --ifm-table-head-background: rgba(0, 0, 0, 0.06);
```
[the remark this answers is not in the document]
slorber: maybe we could have a stronger border-bottom-color ?

as the border width is already larger that might be visible enough?

PR #228, opened 2022-03-25
fix: proper sizing for hr element

lex111:
Now the sizing for `<hr>` element is confusing: although it is set as 1px by default, but it is actually 2px because the top and bottom borders are summed.


slorber:
agree 1px looks better and it's more intuitive to customize 👍

PR #230, opened 2022-03-29
fix: fix flickering dropdown on hover over its corner

lex111:
Should fix https://github.com/facebook/docusaurus/issues/7046


Josh-Cena (approved):
Looks like it does work better 👍

PR #231, opened 2022-03-29
Markdown: Adapt font-size on mobile

Pierre-Gilles:
Following my PR on Docusaurus: https://github.com/facebook/docusaurus/pull/7004#issuecomment-1081214521

The goal here is to fix the size of titles on mobile (too big right now).

I'm not 100% sure of this PR, I'm not an expert in .pcss. 

How could I test this PR on the docusaurus repo ? 

Docusaurus PR

### DIFF
--- a/packages/core/styles/content/markdown.pcss
+++ b/packages/core/styles/content/markdown.pcss
@@ -92,3 +92,17 @@
     }
   }
 }
+
+@media (max-width: 576px) {
+  .markdown {
+    & h1:first-child {
+      --ifm-h1-font-size: 2rem;
+    }
+    & > h2 {
+      --ifm-h2-font-size: 1.5rem;
+    }
+    & > h3 {
+      --ifm-h3-font-size: 1.25rem;
+    }
+  }
+}


lex111:
How about making it more dynamic? For example, calculate the font size of headings with `calc`, based on some "base factor size"? I mean something similar as in Bootstrap.
 https://github.com/twbs/bootstrap/blob/c11c3a7e1941032be9f239520a52e6064af3e017/scss/_variables.scss#L429

Pierre-Gilles:
@lex111 Why not, but I'm really not sure how to do that ! 😄

lex111:
Honestly, I don't know much about adaptive typography, but I would like to see something like this
https://github.com/sloria/www/blob/278e1efd7fe11c6b6b1345106a4e879a6dc700ad/content/2017-12-30-responsive-typography.md#using-the-media-function

Apparently this would require more effort than just adjusting the headings sizing for mobiles, but if you want to try it, that would be great.

slorber:
That looks good enough to me.

Maybe we can ship a simple solution now, and if someone wants to build something more advanced we can do that later?

lex111:
Although adding new magic values is a little scary, I'm fine with this for now, since I have no idea yet how to implement responsive typography properly.

slorber:
ok, let's do this for now then, we'll find a way to improve later

PR #232, opened 2022-04-01
refactor: better positioning of caret icon in menu item

lex111:
Since soon it will be possible to place pure HTML markup, including SVG icons, inside the navbar items, they will now be mis-positioned in the mobile sidebar.

| Before   | After    |
| -------- | -------- |
| image | image |

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -221,7 +221,11 @@
                 <a class="menu__link" href="#url">Tutorial</a>
               </li>
               <li class="menu__list-item">
-                <a class="menu__link menu__link--sublist" href="#url">v2.0</a>
+                <a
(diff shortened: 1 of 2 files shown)


packages/core/demo/index.html
```diff
@@ -221,7 +221,11 @@
                 <a class="menu__link" href="#url">Tutorial</a>
               </li>
               <li class="menu__list-item">
-                <a class="menu__link menu__link--sublist" href="#url">v2.0</a>
+                <a
+                  class="menu__link menu__link--sublist menu__link--sublist-caret"
```
lex111, line 225: Perhaps we should also add a vertical center alignment for caret? This is especially noticeable on multi-line text:

| Before   | After    |
| -------- | -------- |
| image | image |

lex111: Alright, done it.

PR #233, opened 2022-04-10
feat: introduce class for reset list

lex111:
Similar to `clean-btn`, this class is designed to reset the default browser styles, but for lists this time. It is more useful in Docusaurus code base than in Infima, because it makes sense to use it on _infrequent_ elements, such as footer lists, so as not to clutter up the HTML markup too much.


slorber:
👍  fine with that

PR #234, opened 2022-04-14
refactor: make breadcrumbs tighter

lex111:
Currently the spacing of breadcrumbs seems rather large, I suggest making them more compact.

| Before   | After    |
| -------- | -------- |
| image | image |


slorber:
👍

not a huge difference for me but why not

PR #238, opened 2022-04-15
refactor: improve avatar markup/styles

lex111:
Currently the avatar markup seems a bit verbose, as do the styles (duplicate height/width for each sizing), so let's fix it. Visible result will be same as before.

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2300,9 +2300,10 @@ const myFun = (x, y) =&gt; {
             <div class="col">
               <div class="avatar">
                 <a
-                  class="avatar__photo avatar__photo--xl avatar__photo-link"
+                  class="avatar__photo-link"
                   href="https://twitter.com/dan_abramov">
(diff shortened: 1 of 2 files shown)


packages/core/demo/index.html
```diff
@@ -2299,10 +2299,9 @@ <h3>Horizontal Avatars</h3>
           <div class="row">
             <div class="col">
               <div class="avatar">
-                <a
-                  class="avatar__photo avatar__photo--xl avatar__photo-link"
-                  href="https://twitter.com/dan_abramov">
+                <a href="https://twitter.com/dan_abramov">
```
lex111, line 2302: I guess after these changes it will be possible to remove these custom styles in Docusaurus https://github.com/facebook/docusaurus/blob/6cbc58943e42b54067ffd4d314e5f657e115903a/packages/docusaurus-theme-classic/src/theme/BlogPostAuthor/styles.module.css

PR #240, opened 2022-04-19
refactor: use flex for pills spacing

lex111:
A much better approach that also takes less code.


slorber:
For some reason spacing is now smaller, is this on purpose?

CleanShot 2022-04-20 at 14 46 53@2x

vs 

CleanShot 2022-04-20 at 14 47 13@2x


--- 


https://deploy-preview-240--infima.netlify.app/demo/#pills

vs

https://infima.dev/demo/#pills

slorber:
(otherwise I'm fine with the refactor, let's consider flexbox gap has good enough browser support)

PR #241, opened 2022-04-20
refactor: apply hover state on breadcrumb link only

lex111:
The hover state of breadcrumb item should not be triggered when hovering over breadcrumb separator.


slorber:
👍 thanks

PR #242, opened 2022-04-20
refactor: remove useless spacing for tabs

lex111:
This tiny margin only applies on mobiles and only for full-width tabs, so it's not really clear why it's needed, so let's remove it.


slorber:
It has a little impact but minor, nobody will probably even notice 

👍

PR #243, opened 2022-04-23
refactor: remove code duplication in buttons styles

lex111:
Trying a modern way to solve the specificity issue to reduce the final CSS bundle size.


Josh-Cena (commented):
Same concern about compatiblity as #246. Support for `:where` is only 88%. I understand we are already using this in Docusaurus, but that's in a non-critical place. I wonder if it's wise to use this in a CSS lib.

lex111:
Yes, it's modern CSS framework that targets on popular browsers which usually supports relative new features as `where`.

PR #244, opened 2022-04-25
refactor: use grid layout for pagination nav

lex111:
- Less code
- Simpler markup
- No need for adding an empty element if there is no next/prev link

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2163,63 +2163,62 @@ const myFun = (x, y) =&gt; {
           <div class="row">
             <div class="col col--5">
               <nav class="pagination-nav">
-                <div class="pagination-nav__item">
-                  <a class="pagination-nav__link" href="#url">
-                    <div class="pagination-nav__sublabel">Previous</div>
-                    <div class="pagination-nav__label">Installation</div>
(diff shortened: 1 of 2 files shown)


packages/core/demo/index.html
```diff
@@ -2163,63 +2163,62 @@ <h1>
...
           </div>
           <div class="row">
             <div class="col col--5">
               <nav class="pagination-nav">
-                <div class="pagination-nav__item">
-                  <a class="pagination-nav__link" href="#url">
-                    <div class="pagination-nav__sublabel">Previous</div>
-                    <div class="pagination-nav__label">Installation</div>
-                  </a>
-                </div>
```
Josh-Cena, line 2209: Is this change necessary, or is it just an optimization?

Josh-Cena: Oh... I see that `.pagination-nav__item` doesn't exist anymore

lex111: Just a slight improvement.

Josh-Cena:
Noticed a little inconsistency we had previously:

Screen Shot 2022-04-25 at 3 44 10 PM

This change looks like an improvement 👍

PR #245, opened 2022-04-25
refactor: use flex layout for pagination

lex111:
Just a small optimization to simplify the code.

### DIFF
--- a/packages/core/styles/components/pagination.pcss
+++ b/packages/core/styles/components/pagination.pcss
@@ -6,19 +6,18 @@
  */
 
 :root {
-  --ifm-pagination-border-radius: calc(
-    var(--ifm-global-radius) * var(--ifm-pagination-size-multiplier)
-  );
+  --ifm-pagination-border-radius: var(--ifm-global-radius);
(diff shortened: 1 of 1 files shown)


packages/core/styles/components/pagination.pcss
```diff
@@ -6,19 +6,18 @@
...
-  );
+  --ifm-pagination-border-radius: var(--ifm-global-radius);
   --ifm-pagination-color-active: var(--ifm-color-primary);
   --ifm-pagination-font-size: 1rem;
   --ifm-pagination-item-active-background: var(--ifm-hover-overlay);
-  --ifm-pagination-page-spacing: 0.0625rem;
+  --ifm-pagination-page-spacing: 0.2em;
   --ifm-pagination-padding-horizontal: calc(var(--ifm-spacing-horizontal) * 1);
   --ifm-pagination-padding-vertical: calc(var(--ifm-spacing-vertical) * 0.25);
-  --ifm-pagination-size-multiplier: 1;
```
lex111, line 18: Not very useful in this case, so I suggest remove it for now.

Josh-Cena (approved):
Result seems to be the same

PR #246, opened 2022-04-27
refactor: clean up button-group styles

lex111:


packages/core/styles/components/button-group.pcss
```diff
@@ -6,27 +6,23 @@
...
-      margin-left: var(--ifm-button-group-margin);
     }
 
     &:not(:last-child) {
       border-bottom-right-radius: 0;
       border-top-right-radius: 0;
     }
-
-    &--active {
-      z-index: 1; /* Make the active button's border take precedence. */
```
lex111, line 28: Doesn't make sense, don't know why it's needed at all.

Josh-Cena (commented):
Support for `gap` seems inadequate: only 88%. Should we have some policies about what percentage we aim to support? (I also don't know if it can be mitigated through autoprefixing/etc. but from the dist CSS it seems unlikely)

lex111:
It's a pretty big support (eg `margin-top` has 94.5% support, not 100%), all major browsers support it, we already actively use this property in other places, so shouldn't be any problem here.

Josh-Cena:
OK let's do this 👍

`margin-top` not having 100% is because there's unknown support, though, not because it's actually unsupported

PR #247, opened 2022-04-28
refactor: normalize menu link horizontal spacing

lex111:
As a rule, the horizontal spacing should not be more than twice that of the vertical one (eg, this is true for dropdown items).


Josh-Cena (approved):
Thanks! Looks more proportioned this way

PR #248, opened 2022-04-28
refactor: reduce CSS specificity for grid rules

lex111:
There's just no need for it.

### DIFF
--- a/packages/core/styles/layout/grid.pcss
+++ b/packages/core/styles/layout/grid.pcss
@@ -48,7 +48,7 @@
     }
 
     @media (--ifm-narrow-window) {
-      &.col.col {
+      & {
         --ifm-col-width: 100%;
         flex-basis: var(--ifm-col-width);
         margin-left: 0;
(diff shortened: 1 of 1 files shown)


Josh-Cena (commented):
The idea seems to be that in order to use grid columns you have to also specify `col` as a class, not just `col--3` on itself. But I'm fine with dropping `col`. WDYT?

yangshun:
Agreed it's not needed now. Back then I thought it was helpful to have a class to indicate it's a grid `col`, and add modifiers on top of that for the col size. 

Note that if you remove the `col` class from Docusaurus theme components it might break sites relying on this class for additional styling.

lex111:
I would actually also reduce current `.row .col...` selector paths to just `.col`. As I see from other CSS frameworks, it is common practice.

> Note that if you remove the col class from Docusaurus theme components it might break sites relying on this class for additional styling.

It is unlikely there will be something like this in the built-in Docusaurus components, so I suppose it is a safe change..

Josh-Cena:
> there will be something like this in the built-in Docusaurus components

image

We do extensively use `col--x` though, even in the init template. I'm not sure if the goal is to eventually get rid of those `col` classes.

lex111:
Don't know what you mean, I'm just suggesting to shorten the current selectors, it shouldn't break anything.

Josh-Cena:
Ok, so the goal is to _not_ remove the `col` classes from Docusaurus components, only to shorten the selectors. Should be good then.

lex111:
But I still wanted to try to make the selectors even shorter (by dropping `.row`), but you already merged the PR...

Josh-Cena:
No worries. If you want to refactor it further, just open a new one.

PR #249, opened 2022-04-28
fix: correct the HR element display, update demo

Simek:
# Why

This PR fixes the HR element display issue seen on the Docusaurus website, and other sites based on Infima. 

This horizontal rule only display wrongly on some browsers and system (for sure on Chrome/Edge in Windows with scaling on or on high DPI displays), so this might be the reason why it has not been caught earlier.

# How

The PR changes the HR styling from border based to the height based which is a common practice to avoid border related issues. It also means that user now have full control of the thickness, when earlier, border set to, for example, 2px means that the element will be 4px high.

I have also updated the demo app and added HR element in there.


slorber:
looks fine 👍

PR #253, opened 2022-05-02
refactor: make column classes standalone

lex111:
Follow-up of #248


Josh-Cena (approved):
Let's hope we don't get any spacebar heating problems😄

Issue #226, opened 2022-03-23
SCSS rewrite proposal


RDIL:
Copied from Discord:

I think that for Infima, we should possibly consider moving to something like SCSS instead of postcss.

Why:
- Current setup is highly opinionated, and frowned upon by the postcss core team
- It's not as easy to maintain as it probably could be
- It's difficult to build without adding lots of dependencies out of this repository

Pros:
- More widely supported toolchain, compilers for which are available in many different languages, e.g. sass and node-sass
- Could allow for enhanced treeshaking for projects like mine that use ifm outside Docusaurus.
- Better IDE support potentially

Cons:
- Rewriting of some syntax would be required
- It's a decently big shift

Let me know what y'all think.

slorber:
IMHO nothing is blocking us in this repo from providing the value that we aim to provide to our end users. 

Refactoring everything just for the beauty of our setup is not really worth it unless to do so pre-emptively to avoid potential future blockers.

In any case, I'd prefer delaying this refactor after the 2.0 launch to avoid unnecessary distractions on something that could eventually become a time-sink.

> Could allow for enhanced treeshaking for projects like mine that use ifm outside Docusaurus.

That's the first time I hear Infima is used outside Docusaurus. 
I'm not sure it's a great idea personally and wonder what are your motivations.

lex111:
We actually used SASS syntax before, but we originally also used PostCSS, so decided to limit the number of tools to use and get rid of SASS in the end.
Infima is a pretty small project, so PostCSS is very good for it. Maybe in the next major version migrating to SASS would be justified, but definitely not right now.

RDIL:
@slorber I'm using it to design user interfaces that match the documentation which uses Docusaurus.

RDIL:
Closing this as we now have an internal fork of Infima as of alpha 39.

PR #254, opened 2022-05-03
refactor: exclude source files from npm package

lex111:


yangshun:
The idea of including the source files in the published package is so that
if people want to use their own postcss config to bundle, they can. It's
similar to how Bootstrap includes their raw Sass files for people who want
to have more finegrain control over the output.e

Josh-Cena:
PostCSS setup is quite hard to replicate, so I won't expect people to do that and produce output different from ours. Do we have benchmark about how much publish size this will save?

We do use PostCSS in Docusaurus, but only for RTL and autoprefixer...

lex111:
It is doubtful that anyone uses these source files, alright, let's keep them then.

slorber:
Also related to the proposal to rewrite to SCSS by @RDIL: https://github.com/facebookincubator/infima/issues/226

My personal opinion remains that having Infima/Docusaurus being separated increases the public API surface, adds dev friction, and makes it difficult to test changes in Infima + Docusaurus integration in an atomic way (single PR with deploy preview)... 

For me the less Infima internals we expose the better, but I'm not sure everyone agrees on that 🤷‍♂️

PR #258, opened 2022-06-09
fix: change directions of breadcrumb arrows in rtl 

tildemarky:
This rotates the arrow depending on if rtl mode is set in the html or not

rtl breadcrumb

fix: https://github.com/facebook/docusaurus/issues/7565

### DIFF
--- a/packages/core/styles/components/breadcrumb.pcss
+++ b/packages/core/styles/components/breadcrumb.pcss
@@ -19,6 +19,14 @@
   --ifm-breadcrumb-separator-size-multiplier: 1.25;
 }
 
+html:dir(rtl) .breadcrumbs {
+  &__item {
+    &:not(:last-child):after {
+      transform: rotate(180deg);
+    }
+  }
+}
+
 .breadcrumbs {
   margin-bottom: 0;
   padding-left: 0;


Josh-Cena:
```diff
diff --git a/packages/core/dist/css/default/default.css b/packages/core/dist-branch/css/default/default.css
index f9511[7](https://github.com/facebookincubator/infima/runs/6816723951?check_suite_focus=true#step:4:7)b..73[8](https://github.com/facebookincubator/infima/runs/6816723951?check_suite_focus=true#step:4:8)5[9](https://github.com/facebookincubator/infima/runs/6816723951?check_suite_focus=true#step:4:9)60 100644
--- a/packages/core/dist/css/default/default.css
+++ b/packages/core/dist-branch/css/default/default.css
@@ -1673,6 +1673,[10](https://github.com/facebookincubator/infima/runs/6816723951?check_suite_focus=true#step:4:10) @@ hr {
  * LICENSE file in the root directory of this source tree.
  */

+html[dir="rtl"] .breadcrumbs__item:not(:last-child):after {
+      transform: rotate(180deg);
+    }
+
 .breadcrumbs {
   margin-bottom: 0;
   padding-left: 0;
```

Josh-Cena (approved):
LGTM, thanks for your contribution!

Waiting for @slorber to make the stamp

yangshun (approved):

slorber (approved):
LGTM, just a little refactor 👍 

Surprised because our postcss plugin usually can flip rotations (90deg becomes - 90deg) but apparently it does not flip 0 to 180

Note: it might be possible to reduce "default.css" size using rtlcss directives: https://rtlcss.com/learn/usage-guide/control-directives/
(I mean, the rotation rule only really needs to be in the default-rtl.css file, not both files)

slorber:
> Made the requested changes, looks like the raw control directive is the only way to make a style rtl-only, unless I missed something on the rtlcss repo!

Thanks

Of course the comment stays in the ltr file 😅 but not a big deal as it can be optimized. Afaik our CSSNano preset remove those (`discardComments: {removeAll: true},`)

Issue #239, opened 2022-04-16
Logo not loading in 'Multi-Column' example for footer


iammati:
URL: https://infima.dev/docs/components/footer
The logo isn't loading and you guys may want to change the "2019" to something more dynamic in all three examples (Simple, Multi-Column and Dark) :)

PS: Logo also affects to the "Dark" section right below

![image](https://user-images.githubusercontent.com/41418763/163689062-18f29bf6-d687-46de-be73-4b017087d49b.png)

Josh-Cena:
Thanks for the report. The broken image looks worth fixing, but the copyright is by design, because we are only demonstrating static markup.

There are two problems here: (1) we shouldn't use `src="img/fb-oss-dark.png"` because that's relative to the current URL instead of relative to the root (2) the image simply doesn't exist.

@yangshun Do you have the image for `fb-oss-dark.png`? I only have the version meant to be used on dark backgrounds.

slorber:
I don't think I ever saw a variant supposed to be displayed on light background

also received new Meta assets from Joel and there's no asset to be displayed on light background either

This one is likely the one we'll start using in the future:

![Meta_lockup_negative_primary_RGB](https://user-images.githubusercontent.com/749374/164481121-88b89b69-56b6-4fe9-8ef3-7adccac53cec.png)

slorber:
FYI those are all the pngs I got from Joel:
![Meta_lockup_positive_primary_RGB](https://user-images.githubusercontent.com/749374/164481485-14668f0a-a8d9-47cd-bd4c-105ff4b064a5.png)
![Meta_lockup_negative_primary_RGB](https://user-images.githubusercontent.com/749374/164481504-7bd9a883-e54d-4d7c-be1c-ef18ab6c9163.png)
![Meta_lockup_mono_black_RGB](https://user-images.githubusercontent.com/749374/164481514-1c95f3f1-8dbe-4f98-8053-ea6cfec14b0d.png)
![Meta_lockup_mono_white_RGB](https://user-images.githubusercontent.com/749374/164481520-6ac34d91-ceff-42bc-acc3-2ae5af28428d.png)

(Also have SVG/AI)

Josh-Cena:
Geez these images are legit making my browser lag...

iammati:
> Geez these images are legit making my browser lag...

Same here lol you guys should maybe investigate into that since those images shouldn't cause a browser to lag that much at all 🤣
[...]

PR #263, opened 2022-06-23
Fix broken footer logo in Infima Footer doc

slorber:
Fixes https://github.com/facebookincubator/infima/issues/239

Note there's still one case where we display the logo for dark mode on a light background:

CleanShot 2022-06-23 at 10 42 56@2x

I think it's better to not fix it because the documented markup is static (but the theme/backgroundColor is dynamic): fixing this would likely create noise in our doc for little ROI

### DIFF
--- a/website/docs/components/footer.mdx
+++ b/website/docs/components/footer.mdx
@@ -18,7 +18,7 @@ hide_table_of_contents: true
       <span class="footer__link-separator">&middot;</span>
       <a class="footer__link-item" href="#url">Contribute</a>
     </div>
-    <div>Copyright © 2019 Facebook, Inc.</div>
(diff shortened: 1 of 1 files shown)


Josh-Cena:
> Note there's still one case where we display the logo for dark mode on a light background:

Would using another src work?

slorber:
> Would using another src work?

No because then it's the other theme that would be broken 🤪 

CleanShot 2022-06-23 at 10 48 08@2x


We'd need something akin to `<ThemedImage>` but just in static markup: we don't have the concept of themed image in Infima directly. Maybe we should 🤷‍♂️ 

If you find a solution without polluting the documentation with useless technical code, why not. I'd rather not invest too much time on this little thing.

Josh-Cena:
Ah, didn't realize the footer is themed.

PR #266, opened 2022-06-25
fix: breadcrumb links without `href` should not have hover styles

srmagura:
Breadcrumb items do not necessarily have an `href`, but they always change color when the user hovers over them. The hover styles give the impression that the breadcrumb item is a clickable link even if this is not the case.

This can be seen in the [Redux documentation](https://redux.js.org/tutorials/essentials/part-1-overview-concepts) — neither "Tutorials" nor "Redux Essentials" are links, yet they have hover styles.

This PR uses the `:link` pseudoselector to fix this.

Edit: Signed the CLA.

### DIFF
--- a/packages/core/styles/components/breadcrumb.pcss
+++ b/packages/core/styles/components/breadcrumb.pcss
@@ -71,7 +71,7 @@
       );
     @mixin transition background color;
 
-    &:hover {
+    &:link:hover {
       background: var(--ifm-breadcrumb-item-background-active);
       text-decoration: none;
     }
--- a/packages/core/styles/components/breadcrumb.pcss
+++ b/packages/core/styles/components/breadcrumb.pcss
@@ -71,7 +71,7 @@
       );
     @mixin transition background color;
 
-    &:link:hover {
+    &:any-link:hover {
       background: var(--ifm-breadcrumb-item-background-active);
       text-decoration: none;
     }
--- a/packages/core/styles/content/typography.pcss
+++ b/packages/core/styles/content/typography.pcss
@@ -41,7 +41,7 @@ a {
   text-decoration: var(--ifm-link-decoration);
   @mixin transition color;
 
-  &:link:hover {
+  &:any-link:hover {
     color: var(--ifm-link-hover-color);
     /* autoprefixer: ignore next */
     text-decoration: var(--ifm-link-hover-decoration);
--- a/packages/core/styles/content/typography.pcss
+++ b/packages/core/styles/content/typography.pcss
@@ -41,7 +41,7 @@ a {
   text-decoration: var(--ifm-link-decoration);
   @mixin transition color;
 
-  &:hover {
+  &:link:hover {
     color: var(--ifm-link-hover-color);
     /* autoprefixer: ignore next */
     text-decoration: var(--ifm-link-hover-decoration);
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2288,6 +2288,26 @@ const myFun = (x, y) =&gt; {
               </nav>
             </div>
           </div>
+          <div class="row">
+            <div class="col">
+              <nav aria-label="breadcrumbs">
+                <ul class="breadcrumbs breadcrumbs--lg">
+                  <li class="breadcrumbs__item">
+                    <a class="breadcrumbs__link">Breadcrumbs</a>
+                  </li>
+                  <li class="breadcrumbs__item">
+                    <a class="breadcrumbs__link">Without</a>
+                  </li>
+                  <li class="breadcrumbs__item">
+                    <a class="breadcrumbs__link">Href</a>
+                  </li>
+                  <li class="breadcrumbs__item breadcrumbs__item--active">
+                    <a class="breadcrumbs__link">Test</a>
+                  </li>
+                </ul>
+              </nav>
+            </div>
+          </div>
         </div>
         <div class="section">
           <h1>


packages/core/styles/components/breadcrumb.pcss
```diff
@@ -71,7 +71,7 @@
       );
     @mixin transition background color;
 
-    &:hover {
+    &:link:hover {
```
slorber, line 74: > The :link CSS pseudo-class represents an element that has not yet been visited.

Does it mean that we'll lose the hover effect once the link is visited? 😅 

Maybe using `:is(a):hover` would be better? (90% support good enough for this usecase)

TIL there's also an `:any-link` pseudo-selector, looks like a potential good fit?

packages/core/styles/content/typography.pcss
```diff
@@ -41,7 +41,7 @@ a {
   text-decoration: var(--ifm-link-decoration);
   @mixin transition color;
 
-  &:hover {
+  &:link:hover {
```
slorber, line 44: 🤷‍♂️ this one looks more general and not sure it's directly related to your problem, considering the element is already a `<a>` there?

Can it have unwanted side effects? + not sure `:link` is what we want anyway

slorber (changes requested):
Agree this UX is not ideal and we should change it.

> Breadcrumb items do not necessarily have an href, but they always change color when the user hovers over them. 

In practice it's more:

> breadcrumb items might be simple `<span>` instead of `<a>` when we don't have a link

Not 100% sure using `:link` is the solution here?

---

Also worth adding a test to our demo page here: https://deploy-preview-266--infima.netlify.app/demo/

We only have breadcrumb items with links currently (unlike what we have on the Docusaurus side in practice). Please try to add something here that helps reviewing the current PR.

packages/core/styles/content/typography.pcss
srmagura: You can have `<a>` elements that don't represent links (like headings). It usually only makes sense to display hover styles if the `<a>` represents a link.

srmagura:
Thanks for the feedback. I switched to `:any-link` and added an example.

> breadcrumb items might be simple <span> instead of <a> when we don't have a link

True, but I assume you would not want hover styles if the breadcrumb item is a `<span>`, correct?

Josh-Cena:
```diff
  diff --git a/packages/core/dist/css/default/default.css b/packages/core/dist-branch/css/default/default.css
  index 95fc6e3..f258e29 100644
  --- a/packages/core/dist/css/default/default.css
  +++ b/packages/core/dist-branch/css/default/default.css
  @@ -1233,[7](https://github.com/facebookincubator/infima/runs/7197269101?check_suite_focus=true#step:4:7) +1233,25 @@ a {
     transition: color var(--ifm-transition-fast) var(--ifm-transition-timing-default);
   }
  
  -a:hover {
  +a:link:hover, a:visited:hover {
  +    color: var(--ifm-link-hover-color);
  +    /* autoprefixer: ignore next */
  +    text-decoration: var(--ifm-link-hover-decoration);
  +  }
  +
  +a:-webkit-any-link:hover {
  +    color: var(--ifm-link-hover-color);
  +    /* autoprefixer: ignore next */
  +    text-decoration: var(--ifm-link-hover-decoration);
  +  }
  +
  +a:-moz-any-link:hover {
  +    color: var(--ifm-link-hover-color);
  +    /* autoprefixer: ignore next */
  +    text-decoration: var(--ifm-link-hover-decoration);
  +  }
  +
  +a:any-link:hover {
       color: var(--ifm-link-hover-color);
       /* autoprefixer: ignore next */
       text-decoration: var(--ifm-link-hover-decoration);
  @@ -172[8](https://github.com/facebookincubator/infima/runs/7197269101?check_suite_focus=true#step:4:8),7 +1746,22 @@ hr {
       transition-timing-function: var(--ifm-transition-timing-default);
     }
  
  -.breadcrumbs__link:hover {
  +.breadcrumbs__link:link:hover, .breadcrumbs__link:visited:hover, area.breadcrumbs__link[href]:hover {
  +      background: var(--ifm-breadcrumb-item-background-active);
  +      text-decoration: none;
  +    }
  +
  +.breadcrumbs__link:-webkit-any-link:hover {
  +      background: var(--ifm-breadcrumb-item-background-active);
  +      text-decoration: none;
  +    }
  +
  +.breadcrumbs__link:-moz-any-link:hover {
  +      background: var(--ifm-breadcrumb-item-background-active);
  +      text-decoration: none;
  +    }
  +
  +.breadcrumbs__link:any-link:hover {
         background: var(--ifm-breadcrumb-item-background-active);
         text-decoration: none;
       }
  ```

**Comment by srmagura** [comment]
@Josh-Cena Why are those changes necessary? I tested in Chrome, Safari, and Firefox and the vendor prefixes were not necessary.

**Comment by Josh-Cena** [comment]
Ah, no, that's the "post diff comment" workflow's result 😄 It can't be posted on PRs created from forks, so I manually post it so we know how the distributed CSS has changed. See https://github.com/facebookincubator/infima/pull/255#issuecomment-1120486531 for example.

**Comment by srmagura** [comment]
Ooooohhhhh, OK. Do I need to do anything to get that workflow to pass?

**Comment by Josh-Cena** [comment]
No, it's a problem with our workflow plus some permission quirks. It's mostly okay this way.

**Comment by slorber** [comment]
LGTM thanks 👍

File: packages/core/styles/components/breadcrumb.pcss (hunk 1 above)
Comment (slorber) [reply packages/core/styles/components/breadcrumb.pcss]: yes, not very intuitive 😅

File: packages/core/styles/content/typography.pcss [hunk 3]
```diff
@@ -41,7 +41,7 @@ a {
   text-decoration: var(--ifm-link-decoration);
   @mixin transition color;
 
-  &:hover {
+  &:any-link:hover {
     color: var(--ifm-link-hover-color);
     /* autoprefixer: ignore next */
     text-decoration: var(--ifm-link-hover-decoration);
```
pranabdas, line 47: I guess this is causing `text-decoration: underline` for all links including buttons, sidebar menu, navbar, navbar dropdown menu. https://github.com/facebook/docusaurus/issues/7748

yangshun: I think adding this pseudo selector increases the specificity

srmagura: Thanks for pointing this out.

https://github.com/facebookincubator/infima/pull/267

PR #267, opened 2022-07-09
Fix text-decoration regression

srmagura:
Regression introduced here: https://github.com/facebookincubator/infima/pull/266  (adding `:any-link` increases the specificity which prevents other styles from being applied)

Issue in Docusaurus repo: https://github.com/facebook/docusaurus/issues/7748


slorber:
Thanks @srmagura 

I don't like too much this solution, will use mine instead which reverts to previous CSS specific level: https://github.com/facebookincubator/infima/pull/268

PR #268, opened 2022-07-13
fix: revert useless a:any-link:hover change

slorber:
Revert half of https://github.com/facebookincubator/infima/pull/266

`a:any-link:hover` change is not necessary and was based on an artificial usage of `<a className="xyz">` which we don't have in practice (we use `<span>` in such cases)

---

<details>
<summary>Former issue text before reverting</summary>

Fix issue where all links have unexpected underlines:

CleanShot 2022-07-13 at 12 20 00@2x

Docusaurus issue: https://github.com/facebook/docusaurus/issues/7748

See comment https://github.com/facebookincubator/infima/pull/266/files#r917208453

Alternative PR: https://github.com/facebookincubator/infima/pull/267

Support for `:where` is good enough and we already use it anyway: https://caniuse.com/?search=where

It's a good new tool to keep CSS specificity low

</details>

### DIFF
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -340,7 +340,7 @@
               <nav class="navbar">
                 <div class="navbar__inner">
                   <div class="navbar__items">
-                    <a class="navbar__brand">Infima</a>
+                    <a class="navbar__brand" href="#url">Infima</a>
                     <a class="navbar__item navbar__link" href="#url">Docs</a>
                     <a class="navbar__item navbar__link" href="#url">
                       Tutorial
@@ -386,7 +386,7 @@
               <nav class="navbar navbar--dark">
                 <div class="navbar__inner">
                   <div class="navbar__items">
-                    <a class="navbar__brand">
+                    <a class="navbar__brand" href="#url">
                       <span class="navbar__title text--truncate">
                         Infima long long long long long long long long long long
                         long long long long long long long
@@ -456,7 +456,7 @@
                     </div>
                   </div>
                   <div class="navbar__items navbar__items--center">
-                    <a class="navbar__brand">Infima</a>
+                    <a class="navbar__brand" href="#url">Infima</a>
                   </div>
                   <div class="navbar__items navbar__items--right">
                     <div
--- a/packages/core/demo/index.html
+++ b/packages/core/demo/index.html
@@ -2293,16 +2293,18 @@ const myFun = (x, y) =&gt; {
               <nav aria-label="breadcrumbs">
                 <ul class="breadcrumbs breadcrumbs--lg">
                   <li class="breadcrumbs__item">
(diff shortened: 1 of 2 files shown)


Josh-Cena:
Thought about it, but we may expect more complaints like https://github.com/facebook/docusaurus/issues/7617 coming in 😅 I don't have objections to this, but I wonder what the purpose of changing the style for all `a` elements in https://github.com/facebookincubator/infima/pull/266 was?

slorber:
> I don't have objections to this but I wonder what the purpose of changing the style for all a elements in https://github.com/facebookincubator/infima/pull/266 was?

Yes, I think it may not be useful in the first place 😅 

The created demo looks artificial and does not match what we have on the Docusaurus side.

I've run this query on many Docusaurus pages and couldn't find a single usage: `document.querySelectorAll("a:not(:any-link)")` (and ensured the query actually works with a manual test for sure)

@srmagura I'm going to revert this change and adapt the "artificial" example ensure that breadcrumbs without links are not using `<a>` without href

Josh-Cena (approved):

slorber:
> If the [a](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-a-element.html#the-a-element) element has no [href](https://www.w3.org/TR/2011/WD-html5-author-20110809/links.html#attr-hyperlink-href) attribute, then the element [represents](http://www.w3.org/TR/2011/WD-html5-20110525/rendering.html#represents) a placeholder for where a link might otherwise have been placed, if it had been relevant.

https://www.w3.org/TR/2011/WD-html5-author-20110809/the-a-element.html

I guess we can assume we'll never have this case in the future, but may revisit later the introduction of `:where()` if needed, as it still makes sense technically 🤷‍♂️ 

Note we have Infima demo items using `<a>` without links so I'll also clean these up to match Docusaurus HTML structure

Josh-Cena:
Maybe a11y gurus like @bendmyers has more insight on this 😄

We do have the jsx-a11y rule that forbids this case:

> The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/HEAD/docs/rules/anchor-is-valid.md

slorber:
Will merge this for now, looks good enough but let's continue the discussion

Note that we have some cases of `<a href="#">` in Docusaurus too, wonder if those are legit usages

`document.querySelectorAll("a[href='#']")`

CleanShot 2022-07-13 at 13 23 27@2x

srmagura:
Thanks @slorber !!! Seems like I made this change based on the Infima demo site having an example which does not actually occur in practice. Sorry for the regression.

PR #271, opened 2022-07-21
chore: upgrade dependencies

Josh-Cena:


Josh-Cena:
The dist CSS diff looks sus

Josh-Cena:
Looks like the change is expected, since it [used to reorder selectors](https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-pseudo-class-any-link/CHANGELOG.md#715-june-23-2022) but no longer does

PR #279, opened 2022-12-06
Removing button--outline class from button active

### LINKED ISSUE #273: Code sample for "active" Button include "outline" classes
Small thing. I was looking through the [Button](https://infima.dev/docs/components/button) component variations and noticed that the "active" style's code samples include classes present in the "outline" style. For example:

```html
<button class="button button--outline button--active button--primary">
  Primary
</button>
```

Removing the `button--outline` class appears to retain the active style, any reason it should be left in there?

sanketmunot:
Why this PR:
Should fix #273 

What is the issue:
We are using `button--outline` class in documentation for button active, which is ideally not required.

As part of this PR, removed `button--outline` class from Docs & local Infima

Before change:
image

After change:
image


slorber:
Thanks 👍

PR #281, opened 2023-01-07
feat: disable animated transitions by default when `prefers-reduced-motion: reduce`

Mysterious-Dev:
Linked to https://github.com/facebook/docusaurus/issues/8506.

### DIFF
--- a/packages/core/styles/common/variables.pcss
+++ b/packages/core/styles/common/variables.pcss
@@ -152,6 +152,7 @@
 @media (prefers-reduced-motion: reduce) {
   :root {
     --ifm-transition-fast: 0ms;
+    --ifm-transition-slow: 0ms;
   }
 }
 
--- a/packages/core/styles/common/variables.pcss
+++ b/packages/core/styles/common/variables.pcss
@@ -149,4 +149,10 @@
   --ifm-z-index-overlay: 400;
 }
 
+@media (prefers-reduced-motion: reduce) {
+  :root {
+    --ifm-transition-fast: 0ms;
+  }
+}
+
 @custom-media --ifm-narrow-window (max-width: 996px);
--- a/packages/core/styles/components/dropdown.pcss
+++ b/packages/core/styles/components/dropdown.pcss
@@ -57,10 +57,6 @@
     visibility: hidden;
     z-index: var(--ifm-z-index-dropdown);
     @mixin transition opacity transform visibility;
-
-    @media (prefers-reduced-motion) {
-      @mixin transition visibility;
-    }
   }
 
   &__link {
--- a/packages/core/styles/components/dropdown.pcss
+++ b/packages/core/styles/components/dropdown.pcss
@@ -57,6 +57,10 @@
     visibility: hidden;
     z-index: var(--ifm-z-index-dropdown);
     @mixin transition opacity transform visibility;
+
+    @media (prefers-reduced-motion) {
+      @mixin transition visibility;
+    }
   }
 
   &__link {
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -239,11 +239,8 @@ html[data-theme='dark'],
       right: 0;
       top: 0;
       visibility: hidden;
-      @mixin transition opacity visibility, 100ms, ease-in-out;
-
-      @media (prefers-reduced-motion) {
-        @mixin transition visibility, 100ms, ease-in-out;
-      }
+      @mixin transition opacity visibility, var(--ifm-transition-fast),
+        ease-in-out;
     }
 
     &__brand {
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -265,10 +265,6 @@ html[data-theme='dark'],
       transform: translateZ(0);
       transition: transform var(--ifm-transition-fast) ease-in-out;
 
-      @media (prefers-reduced-motion) {
-        transition: none;
-      }
-
       &--show-secondary {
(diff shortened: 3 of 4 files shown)


Mysterious-Dev:
Ready for review. For infima, it's what I can done. As far as the mobile version of the TOC is concerned, this is normally done on the Docusaurus side.

slorber:
Thanks, will review it soon.

Was wondering if maybe we could use this CSS var trick to reduce motion everywhere at once?
https://www.silvestar.codes/articles/using-css-variables-for-reduced-motion-on-a-global-scale/

```css
:root {
--ifm-transition-fast: 100ms;
}

@media (prefers-reduced-motion: reduce) {
  :root {
    --ifm-transition-fast: 0ms;
  }
}
```

I didn't try nor study that solution in depth but that's probably worth exploring so that we avoid spreading the mediaqueries everywhere.

Mysterious-Dev:
Hmm, intresting, I will try this ^^.

Mysterious-Dev:
By the way, can I add somes new variables for this ? Because in css, I have different speed for some transitions.

slorber:
I don't know, if this makes sense we can create "design tokens" for speeds 🤪 I don't know how many different anim speeds we have

Mysterious-Dev:
For modified files, I have 2 different speeds for navbar (one at 250ms and the other to 100ms).

And for dropdown, I have a transition without "speed".

packages/core/styles/components/navbar.pcss
```diff
@@ -217,6 +217,10 @@ html[data-theme='dark'],
     width: var(--ifm-navbar-sidebar-width);
     @mixin transition opacity visibility transform, 250ms, ease-in-out;
 
+    @media (prefers-reduced-motion) {
+      @mixin transition visibility, 250ms, ease-in-out;
+    }
+
```
slorber, line 223: maybe we can simply remove the 250ms? Maybe it wouldn't change much if it was 200ms?

packages/core/styles/components/dropdown.pcss
```diff
@@ -57,6 +57,10 @@
     visibility: hidden;
     z-index: var(--ifm-z-index-dropdown);
     @mixin transition opacity transform visibility;
+
+    @media (prefers-reduced-motion) {
+      @mixin transition visibility;
+    }
```
slorber, line 63: not sure this is needed

apparently we have these styles:

CleanShot 2023-01-20 at 16 18 35@2x

As far as I understand (was there before) we have a mixin that adds the duration automatically

```js
postcssMixins({
      mixins: {
        transition: (mixin, properties, duration, timing) => {
          duration = duration || 'var(--ifm-transition-fast)';
          timing = timing || 'var(--ifm-transition-timing-default)';

          if (properties.includes(' ')) {
            return {
              'transition-property': properties.replace(/\s/g, ', '),
              'transition-duration': duration,
              'transition-timing-function': timing,
            };
          } else {
            return {
              transition: [properties, duration, timing].join(' '),
            };
          }
        },
      },
    }),
```

packages/core/styles/components/navbar.pcss
```diff
@@ -239,6 +243,10 @@ html[data-theme='dark'],
       top: 0;
       visibility: hidden;
       @mixin transition opacity visibility, 100ms, ease-in-out;
+
+      @media (prefers-reduced-motion) {
+        @mixin transition visibility, 100ms, ease-in-out;
+      }
```
slorber, line 249: similarly, why do the backdrop appear before the sidebar?

I would test with the default 200ms anim delay for both and see if it looks good (not sure but worth testing to see if we can simplify)

packages/core/styles/components/dropdown.pcss
Mysterious-Dev: Done.

packages/core/styles/components/navbar.pcss
Mysterious-Dev: Done.

packages/core/styles/components/navbar.pcss
Mysterious-Dev: I change the speed and delete media query. Don't hesitate to tell me if I need to change anything else.

Mysterious-Dev:
Any news ?

packages/core/styles/common/variables.pcss
```diff
@@ -149,4 +149,10 @@
   --ifm-z-index-overlay: 400;
 }
 
+@media (prefers-reduced-motion: reduce) {
+  :root {
+    --ifm-transition-fast: 0ms;
+  }
+}
```
slorber, line 156:
```suggestion
@media (prefers-reduced-motion: reduce) {
  :root {
    --ifm-transition-fast: 0ms;
    --ifm-transition-slow: 0ms;
  }
}
```

Just removing the slow transitions too, although I'm not sure we actually use any 😅  maybe we will someday

slorber (approved):
LGTM thanks 👍 

The new mobile menu backdrop anim timing doesn't look bad so we can keep it

slorber:
will be integrated in https://github.com/facebook/docusaurus/pull/8674

Mysterious-Dev:
Nice ! However, the animation visible on the video below is managed by Docusaurus (with the `willChange` attribute etc...). I'm not sure how to support the media query 😅.

https://user-images.githubusercontent.com/40738104/219637116-147bc2ce-2a53-4962-aa44-18c9a6e6677e.mp4

slorber:
Yes, I added a few fixes on the Docusaurus side as well to handle that

Mysterious-Dev:
Perfect, thanks ^^.

PR #290, opened 2023-09-06
Update typography.pcss

homotechsual:
Add default underline to links for better non-colour dependant compliance with WCAG 1.4.1 (rule id: link-in-text-block)


hichemfantar:
I agree it should be off by default. Users can easily override the css variable.

It doesn't look good in the navbar

image

PR #291, opened 2023-09-29
Fix: pagination-nav height bug on Safari

rbatsenko:
Hey team,

There seems to be an issue with pagination nav items in Docusaurus on Safari, screenshot attached. I first noticed it on mobile (iPhone), but turns out it was Safari-related, no matter the screen size.

Removing `height: 100%` seem to fix it and do no harm on other browsers (at least on Chrome) 🙂 

Screenshot 2023-09-29 at 10 22 34

### DIFF
--- a/packages/core/styles/components/pagination-nav.pcss
+++ b/packages/core/styles/components/pagination-nav.pcss
@@ -23,7 +23,6 @@
     border: 1px solid var(--ifm-color-emphasis-300);
     border-radius: var(--ifm-pagination-nav-border-radius);
     display: block;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


rbatsenko:
Not sure why the property is still there on Netlify deployed preview 🤔

Josh-Cena (commented):
Hi, I can't see an issue with the Docusaurus site:

image

Tested on both desktop and mobile Safari. Could it be your custom CSS or swizzled components? Granted behavior might be the same after this change, we never know what changes each line of CSS could bring to every site :)

slorber:
Going to close this for now because we need to see the problem in practice. Please create a repro.

PR #298, opened 2024-02-22
Semver-compat upgrade of dependencies

zpao:
Just ran `yarn upgrade` to update these. Confirmed website at least worked.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,153 +2,162 @@
 # yarn lockfile v1
 
 
-"@algolia/autocomplete-core@1.7.4":
-  version "1.7.4"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -5,4 +5,4 @@
   publish = "/website/build"
 
 [build.environment]
-  NODE_VERSION = "16"
+  NODE_VERSION = "18"
```
slorber, line 8:
```suggestion
  NODE_VERSION = "20"
```

Should be fine to upgrade directly to Node 20 (current LTS)

Not sure why but Netlify still tries to build with Node 16 despite this PR setting and the site UI setting 😅 


CleanShot 2024-02-23 at 11 42 09


Maybe I'll merge and see if it works.

slorber:
Ah of course there was a `.nvmrc` that Netlify tried to use 😅 
Will remove it, not sure it's important to force devs to use a specific version for dev, it should work fine in both v18 19 20 21

PR #304, opened 2024-04-03
Document text utility classes.

homotechsual:


slorber:
LGTM thanks 👍

slorber:
Note: it's not super important to invest time in document those because sooner or later I'll merge back Infima into theme classic and cleanup everything we don't use

homotechsual:
> Note: it's not super important to invest time in document those because sooner or later I'll merge back Infima into theme classic and cleanup everything we don't use. Infima will become an internal implementation detail rather than a public API recommended to use.

When that happens would we want to move documentation on the theme framework classes to the Docusaurus documentation anyway to make it easier for folks writing custom components/plugins to reuse theme-classic provided classes if they wish?

slorber:
>  to make it easier for folks writing custom components/plugins to reuse theme-classic provided classes

Actually I want to avoid that. If we introduce a `theme-classic-tailwind` later those Infima classes won't exist. Plugins need to be independent, and one plugin should rather not use classes coming from another.

Plugins like search etc should rather use CSS Modules instead of Infima

PR #308, opened 2024-06-01
fix: navbar search input zoom on focus

### LINKED ISSUE #306: bug: search bar input zooms in on focus on iOS
This is because inputs should have a minimum font size of 16px, anything lower will cause a zoom in on iOS.

Discovered this issue when I recently integrated search into [Actix](https://actix.rs/)

https://stackoverflow.com/a/6394497/11334432

https://github.com/facebookincubator/infima/assets/34947993/1f520e46-d993-423c-b99d-d7d6c89b9418

hichemfantar:
closes #306 

This is because inputs should have a minimum font size of 16px, anything lower will cause a zoom in on iOS.

https://stackoverflow.com/a/6394497/11334432

before:

https://github.com/facebookincubator/infima/assets/34947993/1f520e46-d993-423c-b99d-d7d6c89b9418

after:

https://github.com/facebookincubator/infima/assets/34947993/e5124b37-c72e-4bd2-aa54-77f26e81e23c

### DIFF
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -188,7 +188,7 @@ html[data-theme='dark'],
       color: var(--ifm-navbar-search-input-color);
       cursor: text;
       display: inline-block;
-      font-size: 0.9rem;
+      font-size: 1rem;
       height: 2rem;
       padding: 0 0.5rem 0 2.25rem;
       width: 12.5rem;


hichemfantar:
There's an issue with the cla signature, waiting for a fix

hichemfantar:
@slorber can we merge this please?

slorber:
Thanks

As far as I understand this does not really impact Docusaurus because most Docusaurus sites are using Algolia which already uses 1rem

Note that sooner or later I'll deprecate Infima and merge a part of its code back into the Docusaurus theme classic as an implementation detail, so I wouldn't recommend relying too much on it for your own site styling: you'd rather create your own custom CSS for that because things we don't use in Docusaurus core are likely to be removed.

hichemfantar:
@slorber i agree that the plugin shouldn't rely on infima styles as they are prone to change but for now this a good enough fix (the plugin has 27,693 weekly downloads on [npm](https://www.npmjs.com/package/docusaurus-lunr-search))
Also the tailwind change will most likely warrant a major version bump for docusaurus (v4) and the plugin currently has a [peer dep lock](https://github.com/praveenn77/docusaurus-lunr-search/blob/76c49c4214404533b7d5b0e3faa21fc90a05fe82/package.json#L62) to docusaurus<v3 so technically it shouldn't break existing projects if docusaurus gets a v4 release which it should according to semver.

hichemfantar:
@slorber btw do you plan to publish a new release with this fix in order to fix existing websites?

PR #319, opened 2025-01-14
Update text alignment for better rtl support

hichemfantar:
Change text alignment properties to improve consistency across navbar and pagination components.

### DIFF
--- a/packages/core/styles/components/navbar.pcss
+++ b/packages/core/styles/components/navbar.pcss
@@ -281,7 +281,7 @@ html[data-theme='dark'],
       margin: 0 0 0.2rem -0.5rem;
       padding: 0.6rem 1.5rem;
       position: relative;
-      text-align: left;
+      text-align: start;
       top: -0.5rem;
       width: calc(100% + 1rem);
     }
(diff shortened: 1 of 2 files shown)


slorber:
This change feels harmless but also not particularly useful considering our existing postcss rtl plugin already flips all non-logical properties

Does it have any effect in practice?

How does this improve consistency? Afaik we are almost not using logical properties anywhere so I'd say it makes things less consistent by introducing logical properties but only in a single place? 


An example with the navbar

```css
  margin: 0 0 0.2rem -0.5rem;
  padding: 0.6rem 1.5rem;
  text-align: left;
  top: -0.5rem;
```

If we want consistency, it becomes:

```css
  margin-block-start: 0;
  margin-block-end: 0.2rem;
  margin-inline-start: -0.5rem;
  margin-inline-end: 0;
  padding-block: 0.6rem;
  padding-inline: 1.5rem;
  text-align: start; /* Logical equivalent of left */
  inset-block-start: -0.5rem; /* Logical equivalent of top */
```

But then, CSS is heavier than if we just flipped everything with our PostCSS plugin.

hichemfantar:
Alright, it makes sense to only rely on the plugin to do this. closing

PR #338, opened 2026-02-14
fix(core)!: remove `system-ui` from `--ifm-font-family-base`

tats-u:
Fixes #337

This change should be flagged as a breaking change because we cannot guarantee that blindly updating will not cause issues.


Josh-Cena (approved):
:+1:

